PR #2, opened 2019-10-01
Add scalafmt to the project

PatrykRudnicki:


bambuchaAdm:
> Can you add a precheck script at the same time?

@SteveSugden what do you think about precheck sbt command instead of script ?

PatrykRudnicki:
> Can you add a precheck script at the same time?

Yes, of course

SteveSugden:
Not sure what that is.   Just suggesting we do what our other services do

bambuchaAdm:
> Not sure what that is. Just suggesting we do what our other services do

@SteveSugden Main difference would be not `./precheck.sh` but `sbt precheck`

PatrykRudnicki:
Sorry, last PR created headers in all files

SteveSugden:
Headers are good!  Someone had to do it.

PR #5, opened 2019-10-17
CEDS-1820 Add Stride Auth

ed0906:
Can be tested by running `sbt run` and going to `http://localhost:6799/customs-exports-internal-frontend/hello-world`

You'll need "STRIDE_IDP_STUB", "STRIDE_AUTH", "STRIDE_AUTH_FRONTEND" running


app/uk/gov/hmrc/customsexportsinternalfrontend/config/AppConfig.scala
```diff
@@ -17,21 +17,28 @@
...
 
 @Singleton
-class AppConfig @Inject()(config: Configuration, servicesConfig: ServicesConfig) {
+class AppConfig @Inject()(config: Configuration, servicesConfig: ServicesConfig, environment: Environment) {
   private val contactBaseUrl = servicesConfig.baseUrl("contact-frontend")
 
   private val assetsUrl = config.get[String]("assets.url")
-  private val serviceIdentifier = "MyService"
 
+  private val serviceIdentifier = "customs-exports-internal-frontend"
```
PatrykRudnicki: You can inject app name

`@Named("appName") serviceIdentifier: String`

app/uk/gov/hmrc/customsexportsinternalfrontend/config/ErrorHandler.scala
```diff
@@ -22,10 +22,10 @@ import play.api.i18n.MessagesApi
 import play.api.mvc.Request
 import play.twirl.api.Html
 import uk.gov.hmrc.play.bootstrap.http.FrontendErrorHandler
-import uk.gov.hmrc.customsexportsinternalfrontend.views.html.error_template
+import uk.gov.hmrc.customsexportsinternalfrontend.views.html.error
```
PatrykRudnicki: I think name `error_template` was more readable. I know it's lower case, you shouldn't make mistake with Error as class, but still `error_template` is more intuitive for me.

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
```diff
@@ -0,0 +1,72 @@
...
+class AuthenticatedAction @Inject()(
+  appConfig: AppConfig,
+  override val config: Configuration,
+  override val env: Environment,
+  override val authConnector: StrideAuthConnector,
+  mcc: MessagesControllerComponents,
+  unauthorizedPage: unauthorized
+) extends ActionBuilder[AuthenticatedRequest, AnyContent] with AuthorisedFunctions with AuthRedirects {
+
+  override implicit val executionContext: ExecutionContext = mcc.executionContext
```
PatrykRudnicki: Please inject execution context as implicit

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
```diff
@@ -0,0 +1,72 @@
...
+  override val parser: BodyParser[AnyContent] = mcc.parsers.defaultBodyParser
```
PatrykRudnicki: Please Inject parser

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/Verify.scala
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.customsexportsinternalfrontend.controllers.actions
+
+import javax.inject.Inject
+
+class Verify @Inject()(authenticatedAction: AuthenticatedAction) {
```
PatrykRudnicki: What is the point of this class?

app/uk/gov/hmrc/customsexportsinternalfrontend/views/hello_world.scala.html
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  *@
 
-@import uk.gov.hmrc.customsexportsinternalfrontend.views.html._
+@import uk.gov.hmrc.customsexportsinternalfrontend.views.html.templates.main_template
```
PatrykRudnicki: Probably this will causes compilation warnings

project/AppDependencies.scala
```diff
@@ -10,15 +10,16 @@ object AppDependencies {
     "uk.gov.hmrc"             %% "govuk-template"           % "5.42.0-play-26",
     "uk.gov.hmrc"             %% "play-ui"                  % "8.2.0-play-26",
     "uk.gov.hmrc"             %% "bootstrap-play-26"        % "1.1.0"
-  )
+  ) map(_.withSources())
```
PatrykRudnicki: What this will give us?

test/uk/gov/hmrc/customsexportsinternalfrontend/controllers/HelloWorldControllerSpec.scala
```diff
@@ -16,40 +16,31 @@
...
-  private val fakeRequest = FakeRequest("GET", "/")
-
-  private val env = Environment.simple()
-  private val configuration = Configuration.load(env)
+class HelloWorldControllerSpec extends ControllerLayerSpec {
 
-  private val serviceConfig = new ServicesConfig(configuration, new RunMode(configuration, Mode.Dev))
-  private val appConfig = new AppConfig(configuration, serviceConfig)
-  private val helloWorldPage = app.injector.instanceOf[hello_world]
+  private val helloWorldPage = aRealInstanceOf[hello_world]
```
bambuchaAdm: Why we test view and controller toggether? Looks like mix of unit and interaction tests.

app/uk/gov/hmrc/customsexportsinternalfrontend/config/ErrorHandler.scala
ed0906: I dont mind suffixing it with something but I feel that "template" is the wrong word

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
ed0906: This is how its done in our other UIs

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/Verify.scala
ed0906: I knew this would come up...
It's for readability in Controllers. 
We gradually end up with more actions over time, and i'd rather there was just one way to access them (through this class) rather than injecting many actions into each controller

project/AppDependencies.scala
ed0906: Allows us to look into the source code of those libraries

ed0906: Pretty sure its in our other services, it's very useful. Especially when you get a NullPointer or something inside library code. Without this you just see an interface and a message telling you IntelliJ cant find the sources.

ed0906:
> Looks ok to me. I presume this is all how tariffs ops app did it?

Yeh its more or less a direct copy from tariff-classification-frontend with all the role logic pulled out

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
PatrykRudnicki: In backends and movement frontend we're injecting this

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
PatrykRudnicki: I see it now, I added this somewhere, but don't remember where

project/AppDependencies.scala
PatrykRudnicki: `sbt updateClassifiers` probably is similar, but good, let's add this to every service. Nice!

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
ed0906: In controllers yes, but in declarations frontend `AuthAction` we have

```
class AuthActionImpl @Inject()(
  ...
  mcc: MessagesControllerComponents,
  ...
) extends AuthAction with AuthorisedFunctions {

  implicit override val executionContext: ExecutionContext = mcc.executionContext
  override val parser: BodyParser[AnyContent] = mcc.parsers.defaultBodyParser
  ...
}
```

PatrykRudnicki:
```
class AuthActionImpl @Inject()(override val authConnector: AuthConnector, eoriWhitelist: EoriWhitelist, bodyParsers: PlayBodyParsers)(
  implicit override val executionContext: ExecutionContext
) extends AuthAction with AuthorisedFunctions {

  override val parser: BodyParser[AnyContent] = bodyParsers.anyContent
```

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
ed0906: Controllers all inject it if they need it

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
PatrykRudnicki: Approach from movements frontend. Injecting `PlayBodyParsers` recommended in play documentation from 2.6

PatrykRudnicki: You're just using Execution Context in correct and controlled way, and don't need to Inject whole unused MessagesControllerComponents

ed0906: I need MCC to get `Messages` though

bambuchaAdm: Why AuthAction have to know about translations?

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/Verify.scala
bambuchaAdm: Low cohesion in objects is code smell. Same as injecting object just to access their fields.

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
PatrykRudnicki: You're not using Messages here

ed0906: Its needed to build the unauthorizedPage

test/uk/gov/hmrc/customsexportsinternalfrontend/controllers/ControllerLayerSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+import uk.gov.hmrc.customsexportsinternalfrontend.controllers.actions.AuthenticatedAction
+import uk.gov.hmrc.customsexportsinternalfrontend.controllers.exchanges.{AuthenticatedRequest, Operator}
+import uk.gov.hmrc.customsexportsinternalfrontend.views.html.unauthorized
+import uk.gov.hmrc.play.bootstrap.tools.Stubs.stubMessagesControllerComponents
+
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+
+abstract class ControllerLayerSpec
+    extends WordSpec with MustMatchers with GuiceOneAppPerSuite with MockitoSugar with BeforeAndAfterEach {
```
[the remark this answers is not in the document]
ed0906: This test runs in 500-600ms

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
```diff
@@ -0,0 +1,72 @@
...
+    authorised(AuthProviders(PrivilegedApplication)).retrieve(Retrievals.credentials) {
+      _.map { c =>
+        block(AuthenticatedRequest(Operator(c.providerId), request))
+      } getOrElse {
+        logger.debug("Missing credentials")
+        Future.successful(forbidden)
+      }
+    } recover {
+      case _: NoActiveSession =>
+        toStrideLogin(if (appConfig.runningAsDev) s"http://${request.host}${request.uri}" else request.uri)
```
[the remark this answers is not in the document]
ed0906: It's not required elsewhere and depends on the `request`. 
I don't think we'd gain anything by pulling it out

app/uk/gov/hmrc/customsexportsinternalfrontend/config/AppConfig.scala
```diff
@@ -16,22 +16,32 @@
...
+  servicesConfig: ServicesConfig,
+  environment: Environment,
+  @Named("appName") serviceIdentifier: String
+) {
   private val contactBaseUrl = servicesConfig.baseUrl("contact-frontend")
 
   private val assetsUrl = config.get[String]("assets.url")
-  private val serviceIdentifier = "MyService"
 
+  val runningAsDev: Boolean = {
```
PauloPortugal: why  do we need this to run as a dev?

PauloPortugal: Is it to send the user to the stride auth stub?

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
```diff
@@ -0,0 +1,72 @@
...
+  override val parser: BodyParser[AnyContent] = mcc.parsers.defaultBodyParser
+
+  private val logger = Logger(classOf[AuthenticatedAction])
+
+  override def invokeBlock[A](request: Request[A], block: AuthenticatedRequest[A] => Future[Result]): Future[Result] = {
+    implicit val hc: HeaderCarrier =
+      HeaderCarrierConverter.fromHeadersAndSession(request.headers, Some(request.session))
+    lazy val forbidden = Forbidden(unauthorizedPage()(request, mcc.messagesApi.preferred(request)))
+
+    authorised(AuthProviders(PrivilegedApplication)).retrieve(Retrievals.credentials) {
```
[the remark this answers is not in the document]
PauloPortugal: Looks good to me

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
```diff
@@ -0,0 +1,72 @@
...
+
+@Singleton
+class AuthenticatedAction @Inject()(
+  appConfig: AppConfig,
+  override val config: Configuration,
+  override val env: Environment,
+  override val authConnector: StrideAuthConnector,
+  mcc: MessagesControllerComponents,
+  unauthorizedPage: unauthorized
+) extends ActionBuilder[AuthenticatedRequest, AnyContent] with AuthorisedFunctions with AuthRedirects {
```
PatrykRudnicki: IMO unauthorisedPage should be handled not in the action but in the controller.

MessagesControllerComponents is for Controller not for Action. It's necessary for FrontendController.

IMO This AutheticatedAction is wrong, is not doing what expected. Doing more things than should.

app/uk/gov/hmrc/customsexportsinternalfrontend/config/AppConfig.scala
ed0906: When running locally e.g service manager it needs to redirect to `localhost:someport/path`
When running in dev/qa etc it needs to redirect to `/path` 
So we need to be able to tell it which to redirect to

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
ed0906: What would you suggest?

PatrykRudnicki: Logic and view in Controller/Handler etc, The same approach like we have in other services. Above examples for movements and exports. Movements are closer to correct implementation if you look for the approaches in new play version. (play 2.6 and newer)


Example of unauthorised usage:

            Future.successful(Results.Redirect(routes.UnauthorisedController.onPageLoad()

Controller where we will have logic around unauthorised. If you will have another place where you need to go to unauthorised page with this approach you still need to inject view and everything. 

Unauthorised is treated as normal endpoint in every our service. Implementation in this PR is not based on our services.

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/Verify.scala
ed0906: Removed

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
ed0906: The thing is the UnauthorisedController in those services is just a page with an error message, it doesn't actually mean you are unauthorised. I could be logged in and go to `/unauthorised` for whatever reason and it would incorrectly tell me I am not unauthorised. A little weird.

The way I've implemented it more closely mirrors how the `error` view is used:
- If you access an endpoint that errors you get the error view (not a generic redirect)
Similarly with the way this is implemented:
- If you access an endpoint that you are not allowed to you get an unauthorised view (not a generic redirect)

I don't think there is anything wrong with Actions loading views, it saves us having unnecessary endpoints with generic error messages

PatrykRudnicki: If you go to unauthorised you're expecting that you'll have something related with unauthorised.

In my opinion Action and Controller are two different things. Everyone has different approaches. You can't change my thinking about Action, Controllers. I said everything that I can around those two things. Knowledge and approach based on the documentation and another our services. Still will repeat it, I'm not blocking this, but I won't approve it too.

This discussion doesn't make sense in my opinion, it's going to the wrong direction. It's not necessary :)

If team is happy stay with this, I'm not blocking it :)

app/uk/gov/hmrc/customsexportsinternalfrontend/config/AppConfig.scala
PauloPortugal: Sorry!!!! I should have approved the PR.
You could have delegated that to the config.
All good!

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
bambuchaAdm: So maybe inject it to unauthorized page ?

app/uk/gov/hmrc/customsexportsinternalfrontend/controllers/actions/AuthenticatedAction.scala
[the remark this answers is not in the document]
bambuchaAdm: For reach request we build same - `AuthProviders(PrivilegedApplication)` which look exactly same and does not perform side effect - we can reuse them between requests

test/uk/gov/hmrc/customsexportsinternalfrontend/controllers/HelloWorldControllerSpec.scala
bambuchaAdm: Calling view is what controller does - you can use mock here. Counter prof  by [example](https://github.com/hmrc/customs-declare-exports-frontend/blob/91a50967a6eecf9b7d5a9281a89cc149d855231a/test/unit/controllers/declaration/PreviousDocumentsControllerSpec.scala#L37)

PR #11, opened 2019-10-25
CEDS-1862 Basic support for repository, action etc. and first round of tests

PatrykRudnicki:


app/controllers/ChoiceController.scala
```diff
@@ -19,21 +19,66 @@ package controllers
...
+  getJourney: JourneyRefiner,
+  mcc: MessagesControllerComponents,
+  movementRepository: MovementRepository,
+  choicePage: choice_page
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport {
 
   def displayChoiceForm(): Action[AnyContent] = authenticate.async { implicit request =>
-    Future.successful(Ok(choicePage(Choice.form)))
+  println("#### " + request.operator.pid)
```
ed0906: Presume we don't need these any more?

app/controllers/ChoiceController.scala
```diff
@@ -19,21 +19,66 @@ package controllers
...
+    movementRepository.findByPid(request.operator.pid).map {
+      case Some(cache) =>
+        Ok(choicePage(Choice.form().fill(Choice(cache.answers.`type`))))
+      case None        =>
+        Ok(choicePage(Choice.form()))
+    }.recover {
+      case error: Throwable =>
+        println("####### " + error)
+        println("####### " + error.getCause)
```
ed0906: Or these

app/controllers/ChoiceController.scala
```diff
@@ -19,21 +19,66 @@ package controllers
...
 
   def displayChoiceForm(): Action[AnyContent] = authenticate.async { implicit request =>
-    Future.successful(Ok(choicePage(Choice.form)))
+  println("#### " + request.operator.pid)
+    movementRepository.findByPid(request.operator.pid).map {
+      case Some(cache) =>
+        Ok(choicePage(Choice.form().fill(Choice(cache.answers.`type`))))
+      case None        =>
+        Ok(choicePage(Choice.form()))
+    }.recover {
```
ed0906: I'd let this bubble up to the error handler

conf/application.conf
```diff
@@ -80,18 +83,22 @@ microservice {
...
-  enabled = true
+  showSamples = false
+  jvm = false
+  enabled = false
 }
 
 auditing {
-  enabled = true
-  traceRequests = true
+  enabled = false
```
ed0906: We could do with noting somewhere that we need to override these in  app-config-base

app/controllers/ChoiceController.scala
PatrykRudnicki: It was just for testing

conf/application.conf
PatrykRudnicki: Your propose is to change it to previous implementation? I mean enable metrics and auditing locally?

ed0906: as far as i'm aware, auditing is the only one we need to turn off locally. I'd look at the config for other frontends

PR #15, opened 2019-10-28
CEDS-1825 - SubmissionService impl (no tests)

SteveSugden:


SteveSugden:
Was copied and re-factored.   Lots of bits missing but I can try to copy over some tests.

ed0906:
I think given these components are probably going to be the same in this service we should copy across tests too

PR #16, opened 2019-10-28
CEDS-1862 Associate journey without tests

PatrykRudnicki:


app/controllers/consolidations/AssociateUCRSummaryController.scala
```diff
@@ -0,0 +1,64 @@
...
+    extends FrontendController(mcc) with I18nSupport {
+
+  def displayPage(): Action[AnyContent] = (authenticate andThen getJourney) { implicit request =>
+    val answers = request.answersAs[AssociateUcrAnswers]
+    val mucrOpt = answers.mucrOptions
+    val associateUcrOpt = answers.associateUcr
+
+    (mucrOpt, associateUcrOpt) match {
+      case (Some(mucrOptions), Some(ucr)) => Ok(associateUcrSummaryPage(ucr, mucrOptions.mucr))
+      case _                              => Redirect(controllers.routes.ChoiceController.displayPage())
```
ed0906: There's that ReturnToStartException if you'd prefer to use that

To me, it would make it clearer this is an error scenario

app/controllers/consolidations/AssociateUCRSummaryController.scala
```diff
@@ -0,0 +1,64 @@
...
+  getJourney: JourneyRefiner,
+  mcc: MessagesControllerComponents,
+  movementRepository: MovementRepository,
+  associateUcrSummaryPage: associate_ucr_summary
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport {
+
+  def displayPage(): Action[AnyContent] = (authenticate andThen getJourney) { implicit request =>
+    val answers = request.answersAs[AssociateUcrAnswers]
+    val mucrOpt = answers.mucrOptions
```
ed0906: These two lines could just be `answers.{fieldName}.orElse(throw ReturnToStartException)`

PatrykRudnicki: `orElse` return Option, so if it'll be Some(something) still we will have Option to unwrap. I think current apporach is okay, but I'll add this ReturnToStartException

ed0906: Sorry I meant `answers.{fieldName}.getOrElse(throw ReturnToStartException)`

PatrykRudnicki: Changed

PR #22, opened 2019-10-29
CEDS-1863 Disassociate DUCR fixes

ed0906:


app/services/audit/AuditService.scala
```diff
@@ -117,7 +117,7 @@ class AuditService @Inject()(connector: AuditConnector, @Named("appName") appNam
 
   def getMovementsData(answers: Answers): JsObject = {
 
-    val userInput = Map(
+    val userInput = Map[String, String](
```
ed0906: Json.toJson below doesn't work without this

PR #29, opened 2019-10-30
CEDS-1862 Submitting Associate Request

PatrykRudnicki:


app/controllers/consolidations/AssociateUCRConfirmationController.scala
```diff
@@ -20,17 +20,22 @@ import controllers.actions.{AuthenticatedAction, JourneyRefiner}
...
+  movementRepository: MovementRepository,
   associateUCRConfirmPage: associate_ucr_confirmation
-) extends FrontendController(mcc) with I18nSupport {
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport {
 
-  def displayPage(): Action[AnyContent] = (authenticate andThen getJourney) { implicit request =>
-    Ok(associateUCRConfirmPage())
+  def displayPage(): Action[AnyContent] = (authenticate andThen getJourney).async { implicit request =>
+    movementRepository.removeByPid(request.pid).map(_ => Ok(associateUCRConfirmPage()))
```
ed0906: I'd argue the repository code should be part of the SubmissionService

app/services/SubmissionService.scala
```diff
@@ -54,6 +54,21 @@ class SubmissionService @Inject()(
       }
   }
 
+  def submit(pid: String, answers: AssociateUcrAnswers)(implicit hc: HeaderCarrier): Future[Unit] = {
+    val eori = answers.eori.getOrElse(throw ReturnToStartException)
+    val mucr = answers.mucrOptions.map(_.mucr).getOrElse(throw ReturnToStartException)
+    val ucr = answers.associateUcr.map(_.ucr).getOrElse(throw ReturnToStartException)
+
+    connector
+      .submit(AssociateUCRRequest(pid, eori, mucr, ucr))
```
ed0906: We should probably clear the answers *after* the submission is successful, that way we can resubmit if it fails

app/controllers/consolidations/AssociateUCRConfirmationController.scala
PatrykRudnicki: So in that case you can access confirmation page because you have empty cache.

I think that we can just use confirmation page in Summary Controller without separte endpoint or controller for it...

app/services/SubmissionService.scala
PatrykRudnicki: As above, you will be redirected to the choice page, because cache will be empty. Disassociation doesn't work

PatrykRudnicki: Now when you go through disassociation journey you'll not reach confirmation page, but choice page

app/models/cache/Answers.scala
```diff
@@ -51,7 +51,7 @@ object ArrivalAnswers {
 }
 
 case class AssociateUcrAnswers(
-  override val eori: Option[String] = None,
+  override val eori: Option[String] = Answers.fakeEORI,
```
SteveSugden: Does this work?  Is the fake eori persisted/submitted at the end of the journey?
I think I tried something like this and it didn't work

PatrykRudnicki: The same like passing in Choice Controller as arguments

app/controllers/consolidations/AssociateUCRConfirmationController.scala
SteveSugden: Are you saying that we don't need this controller?  That the AssociateUCRSummaryController should just display the confirmation page (after the service has removed the cache)?

That's what the movement SummaryController does so it would be good to be consistent.  (The Summary controller also calls 'remove' on the repo - so maybe that should move to the service after successful submit as well.

app/models/cache/Answers.scala
SteveSugden: OK - I trust you!

app/controllers/consolidations/AssociateUCRConfirmationController.scala
PatrykRudnicki: Yeah, I added sticker to tech board to display confirmation directly in submit method in Summary page, without creating Confirmation Controller

SteveSugden: So are going to do that now for this service and port solution back to movements?  Or do both as tech debt?

I don't mind.

app/controllers/consolidations/AssociateUCRController.scala
```diff
@@ -45,26 +46,29 @@ class AssociateUCRController @Inject()(
...
+              Redirect(consolidationRoutes.AssociateUCRSummaryController.displayPage())
+            }
           }
-        }
-      )
+        )
+    }
   }
+
+  private def processForm() = ???
```
SteveSugden: Is this used ???

PatrykRudnicki: 🤦‍♂ Thanks, let's mi finish this PR before merging, because as I see I can't prepare something without errors. I'm on it

app/controllers/consolidations/AssociateUCRController.scala
```diff
@@ -45,26 +46,29 @@ class AssociateUCRController @Inject()(
...
-            case None       => Future.successful(Redirect(controllers.routes.ChoiceController.displayPage()))
-          }
-        },
-        formData => {
-          val updatedCache = request.answersAs[AssociateUcrAnswers].copy(associateUcr = Some(formData))
-          movementRepository.upsert(Cache(request.pid, updatedCache)).map { _ =>
-            Redirect(consolidationRoutes.AssociateUCRSummaryController.displayPage())
+    val mucrOptions = request.answersAs[AssociateUcrAnswers].mucrOptions
+
+    if(mucrOptions.isEmpty) throw ReturnToStartException
```
SteveSugden: How about a method that extracts the mucrOptions from the request (or throws the ReturnException).  You could use it below where you need it (to re-populate the form if it fails validation) and in the display method above

app/controllers/consolidations/AssociateUCRController.scala
PatrykRudnicki: Finished

app/controllers/consolidations/AssociateUCRController.scala
SteveSugden: Does look better.   But there is still a "mucrOptions.get" below.

How about `processForm(mucrOptions.getOrElse(throw ReturnToStartException))`
Then processForm does not have to take an Option

app/services/SubmissionService.scala
```diff
@@ -54,6 +54,22 @@ class SubmissionService @Inject()(
...
+      .flatMap(_ => movementRepository.removeByPid(pid))
```
SteveSugden: 👍

PR #31, opened 2019-10-30
CEDS-1825 - movements controller specs

SteveSugden:


test/controllers/movements/SummaryControllerSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+
+    "return 200 (OK)" when {
+
+      "GET displayPage is invoked without data in cache" in {
+
+        givenTheCacheIsEmpty()
+
+        val result = controller().displayPage()(getRequest)
+
+        status(result) mustBe OK
```
ed0906: Can we assert on what view was sent in the response please :)

test/repository/MockCache.scala
```diff
@@ -50,6 +50,9 @@ trait MockCache extends MockitoSugar with BeforeAndAfterEach {
   protected def givenTheCacheIsEmpty(): Unit =
     given(cache.findByPid(any())).willReturn(Future.successful(None))
 
+  protected def givenTheCacheIsCleared(): Unit =
```
ed0906: Dont think we need a method for this, we can just stick it in the `beforeEach`

SteveSugden: Done

test/controllers/movements/SummaryControllerSpec.scala
SteveSugden: Done

PR #37, opened 2019-10-31
CEDS-1863 Disassociate a MUCR/DUCR

ed0906:


app/services/SubmissionService.scala
```diff
@@ -41,7 +41,7 @@ class SubmissionService @Inject()(
 
   def submit(pid: String, answers: DisassociateUcrAnswers)(implicit hc: HeaderCarrier): Future[Unit] = {
     val eori = answers.eori.getOrElse(throw ReturnToStartException)
-    val ucr = answers.ucr.getOrElse(throw ReturnToStartException)
+    val ucr = answers.ucr.getOrElse(throw ReturnToStartException).ucr
 
     connector
       .submit(DisassociateDUCRRequest(pid, eori, ucr))
```
PatrykRudnicki: After submit you're clearing the cache, with this you'll never reach confirmation page, becasue we're using `getJourney` on the confirmation page, which will throw ReturnToStartPage exception.

app/controllers/ChoiceController.scala
```diff
@@ -56,7 +56,7 @@ class ChoiceController @Inject()(
       case forms.Choice.AssociateUCR =>
         proceedJourney(AssociateUcrAnswers(), consolidations.routes.MucrOptionsController.displayPage())
       case forms.Choice.DisassociateUCR =>
-        proceedJourney(DisassociateUcrAnswers(), consolidations.routes.DisassociateDucrController.display())
+        proceedJourney(DisassociateUcrAnswers(), controllers.consolidations.routes.DisassociateUcrController.display())
```
SteveSugden: Can do `consolidations.routes.DisassociateUcrController.display()` (i.e. drop the `controllers`.  But this might be fixed when you update from master

app/controllers/consolidations/DisassociateUcrSummaryController.scala
```diff
@@ -0,0 +1,61 @@
...
+      case _         => throw ReturnToStartException
+    }
+  }
+
+  def submit: Action[AnyContent] = (authenticate andThen getJourney(JourneyType.DISSOCIATE_UCR)).async { implicit request =>
+    val answers = request.answersAs[DisassociateUcrAnswers]
+
+    submissionService.submit(request.pid, answers).map { _ =>
+      Redirect(controllers.consolidations.routes.DisassociateUcrConfirmationController.display())
+        .flashing(FlashKeys.UCR -> answers.ucr.map(_.ucr).getOrElse(throw ReturnToStartException))
```
SteveSugden: Just spotted this - if the ucr is missing then it may be too late at this point - the submission might have happened.  Other controllers are extracting required values before calling the submission service.

app/services/SubmissionService.scala
ed0906: Good spot

app/controllers/consolidations/DisassociateUcrSummaryController.scala
ed0906: Refactored that in my last commit

PR #38, opened 2019-11-01
CEDS-1912 Refactor summary controllers for consistency

SteveSugden:
This is a POC PR for the AssociateUcr journey.   Looks like more changes than there are really as I've also renamed the 2 existing controllers.  The important bits are

- AssociateUcrConfirmationController which attempts to extract the confirmation data from Flash, redirecting to Choice page if it fails (i.e. user has refreshed confirmation page)

- AssociateUcrSummaryController which now redirects to AssociateUcrConfirmationController after submission (rather than rendering the confirmation screen)

### DIFF
--- a/app/controllers/consolidations/AssociateUcrConfirmationController.scala
+++ b/app/controllers/consolidations/AssociateUcrConfirmationController.scala
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2019 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package controllers.consolidations
+
+import controllers.actions.{AuthenticatedAction, JourneyRefiner}
+import controllers.storage.FlashKeys
+import javax.inject.Inject
+import play.api.i18n.I18nSupport
+import play.api.mvc._
+import services.SubmissionService
+import uk.gov.hmrc.play.bootstrap.controller.FrontendController
+import views.html.associate_ucr_confirmation
+
+import scala.concurrent.ExecutionContext
+
+class AssociateUcrConfirmationController @Inject()(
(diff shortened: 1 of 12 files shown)


app/controllers/consolidations/AssociateUcrConfirmationController.scala
```diff
@@ -0,0 +1,45 @@
...
+  authenticate: AuthenticatedAction,
+  mcc: MessagesControllerComponents,
+  associateUCRConfirmPage: associate_ucr_confirmation
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport {
+
+  def display: Action[AnyContent] = authenticate { implicit request =>
+    (request2flash.get(FlashKeys.CONSOLIDATION_KIND), request2flash.get(FlashKeys.UCR)) match {
+      case (Some(kind), Some(ucr)) => Ok(associateUCRConfirmPage(kind, ucr))
+      case _                       => Redirect(controllers.routes.ChoiceController.displayPage)
```
PatrykRudnicki: Instead of manual redirect I'll throw an ReturnToStartPage exception

SteveSugden: That leaves the user looking at a "Something went wrong" screen.
Unless there is a plan to make throwing that exception actually "return to start page"

PatrykRudnicki: Return to start page redirect to Choice page. Is it works in different way now?

ed0906: ReturnToStartException *should* redirect you to Choice. We just need to add that catch & redirect logic into the error handler

app/controllers/consolidations/AssociateUcrConfirmationController.scala
```diff
@@ -0,0 +1,45 @@
...
+
+class AssociateUcrConfirmationController @Inject()(
+  authenticate: AuthenticatedAction,
+  mcc: MessagesControllerComponents,
+  associateUCRConfirmPage: associate_ucr_confirmation
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport {
+
+  def display: Action[AnyContent] = authenticate { implicit request =>
+    (request2flash.get(FlashKeys.CONSOLIDATION_KIND), request2flash.get(FlashKeys.UCR)) match {
```
ed0906: Do we need to change the other Confirmation Controllers to throw ReturnToStart if the Flash data is missing?
Or have they been done already?

SteveSugden: From memory they just display "-" when the data is missing.  This is kind of the aim of this ticket - to get a consistent approach.

PR #40, opened 2019-11-01
CEDS-1912 - store summary data in the cache

SteveSugden:


app/controllers/consolidations/AssociateUcrConfirmationController.scala
```diff
@@ -19,27 +19,25 @@ package controllers.consolidations
...
 
-  def display: Action[AnyContent] = authenticate { implicit request =>
-    (request2flash.get(FlashKeys.CONSOLIDATION_KIND), request2flash.get(FlashKeys.UCR)) match {
-      case (Some(kind), Some(ucr)) => Ok(associateUCRConfirmPage(kind, ucr))
-      case _                       => Redirect(controllers.routes.ChoiceController.displayPage)
-    }
-
+  def display: Action[AnyContent] = (authenticate andThen getJourney) { implicit request =>
+    val summary = request.answers.summary.getOrElse(throw ReturnToStartException)
+    Ok(associateUCRConfirmPage(summary.get(FlashKeys.CONSOLIDATION_KIND), summary.get(FlashKeys.UCR)))
```
PatrykRudnicki: We don't need Flash if we store information in database

PatrykRudnicki: I'm not sure about this. I think easier and better is just clear cache when user is going through Choice page, this will clear old journeys, but it will cause more problems.

SteveSugden: Not using Flash.  Can rename "FlashKeys" constants if we go with this approach.  

But you've given me yet another idea

app/controllers/consolidations/AssociateUcrConfirmationController.scala
```diff
@@ -19,27 +19,25 @@ package controllers.consolidations
...
 )(implicit ec: ExecutionContext)
     extends FrontendController(mcc) with I18nSupport {
 
-  def display: Action[AnyContent] = authenticate { implicit request =>
-    (request2flash.get(FlashKeys.CONSOLIDATION_KIND), request2flash.get(FlashKeys.UCR)) match {
-      case (Some(kind), Some(ucr)) => Ok(associateUCRConfirmPage(kind, ucr))
-      case _                       => Redirect(controllers.routes.ChoiceController.displayPage)
-    }
-
+  def display: Action[AnyContent] = (authenticate andThen getJourney) { implicit request =>
```
ed0906: Interesting approach, If we go with this I think we could create a `getSummary` action which would be used instead of `getJourney` in summary pages.
It could create a 
```
SummaryRequest[T](answers: Map[String, String], request: AuthenticatedRequest[T]) extends WrappedRequest[T](request){
    def get(key: String): String = answers.getOrElse(throw ReturnToStartException)
}
```

SteveSugden: Thanks.  Only worth pursuing if product team care that the summary data is lost after a screen refresh and/or that the 'choice' is not preserved when starting the next journey

ed0906: I think we can preserve or not preserve choice with either solution

PR #41, opened 2019-11-01
CEDS-1860 notifications journey

MaciejRewera:
This PR introduces View Submissions page together with all dependencies in connector and models.


app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -62,4 +64,42 @@ class CustomsDeclareExportsMovementsConnector @Inject()(appConfig: AppConfig, ht
 
   private def logFailedExchange(`type`: String, exception: Throwable): Unit =
     logger.warn(`type` + " failed", exception)
+
+  def fetchAllSubmissions(providerId: String)(implicit hc: HeaderCarrier): Future[Seq[SubmissionFrontendModel]] =
```
ed0906: I think this service calls it `pid` rather than `providerId`, it'd be good to be consistent otherwise we'll look back in the future and forget they're the same thing

app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -62,4 +64,42 @@ class CustomsDeclareExportsMovementsConnector @Inject()(appConfig: AppConfig, ht
...
+    httpClient
+      .GET[Seq[SubmissionFrontendModel]](s"$CustomsDeclareExportsMovementsUrl$FetchAllSubmissionsEndpoint", providerIdQueryParam(providerId))
+      .andThen {
+        case Success(response)  => logger.debug(s"Submissions fetch response. $response")
```
ed0906: Why not use the log methods the other connector methods use?

The intention was for it to be consistent to Kibana searches are easier

app/controllers/ViewSubmissionsController.scala
```diff
@@ -0,0 +1,57 @@
...
+  viewSubmissionsPage: view_submissions
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc) with I18nSupport {
+
+  def displayPage: Action[AnyContent] = authenticate.async { implicit request =>
+    val providerId = request.pid
+
+    for {
+      submissions <- connector.fetchAllSubmissions(providerId)
+      notifications <- Future.sequence(submissions.map(submission => connector.fetchNotifications(submission.conversationId, providerId)))
```
ed0906: This seems inefficient

For many submissions we are making many back end calls.

It'd be better just to query the back end `/notifications` with `providerId`/`pid` and then group by actionID

app/models/notifications/Entry.scala
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.notifications
+
+import models.UcrBlock
+import play.api.libs.json.Json
+
+case class Entry(ucrBlock: Option[UcrBlock] = None, goodsItem: Seq[GoodsItem] = Seq.empty, entryStatus: Option[EntryStatus] = None) {
```
ed0906: What's an "Entry"? Is there a better name we can think of?

app/models/notifications/NotificationFrontendModel.scala
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package models.notifications
+
+import java.time.Instant
+
+import play.api.libs.json.Json
+
+case class NotificationFrontendModel(
```
ed0906: Can we  just call it `Notification`?

app/models/notifications/Entry.scala
PatrykRudnicki: As I remember it's element from schema.

app/views/components/elements_in_separate_lines.scala.html
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *@
+
+@(list: Seq[String])
+
+@for(elem <- list) {
```
ed0906: I'm not convinced we really need a component for this

`for` loops are much more intuitive to use and easy to remember then our own component

app/views/view_submissions.scala.html
```diff
@@ -0,0 +1,67 @@
...
+@main_template(title = Title(messages("submissions.title")), fullWidth = true) {
+
+  @components.back_link(routes.ChoiceController.displayPage())
+
+  @components.page_title(Some("submissions.title"))
+
+  <div class="govuk-grid-column-two-thirds">
+    <p>@messages("submissions.summary") </p>
+    <table>
+      <thead class="govuk-table__head">
```
ed0906: These `govuk` classes do nothing, we should get rid to avoid visual clutter

conf/app.routes
```diff
@@ -60,3 +60,6 @@ POST        /shut-mucr                             controllers.consolidations.Sh
 # Shut a MUCR summary page
 GET         /shut-mucr-summary                     controllers.consolidations.ShutMucrSummaryController.displayPage()
 POST        /shut-mucr-summary                     controllers.consolidations.ShutMucrSummaryController.submit()
+
+# Submissions page
+GET         /movements                             controllers.ViewSubmissionsController.displayPage()
```
ed0906: The submissions page shows consolidations too though doesnt it?
Should it not just be `/submissions`?

test/base/MockExportsMovementsConnector.scala
```diff
@@ -0,0 +1,55 @@
...
+import models.requests.MovementRequest
+import models.submissions.SubmissionFrontendModel
+import org.mockito.ArgumentMatchers.any
+import org.mockito.Mockito._
+import org.scalatest.{BeforeAndAfterEach, Suite}
+import org.scalatestplus.mockito.MockitoSugar
+
+import scala.concurrent.Future
+
+trait MockExportsMovementsConnector extends MockitoSugar with BeforeAndAfterEach { self: Suite =>
```
ed0906: Do we really need this?

Writing Mockito mocks is intuitive. Easy to remember.

I find these classes just make writing tests more complex as you have to remember a bunch of helper methods exist that just make the code less readable (as people forget what they do so use them when they aren't required)

test/connectors/ConnectorSpec.scala
```diff
@@ -16,28 +16,19 @@
...
-  override def fakeApplication(): Application = {
-    SharedMetricRegistries.clear()
-    new GuiceApplicationBuilder().configure(overrideConfig).build()
-  }
+  def overrideConfig: Map[String, Any] = Map()
 
   protected implicit val ec: ExecutionContext = scala.concurrent.ExecutionContext.Implicits.global
   protected implicit val hc: HeaderCarrier = HeaderCarrier()
-  protected val httpClient: DefaultHttpClient = app.injector.instanceOf[DefaultHttpClient]
+  protected val httpClient: DefaultHttpClient = instanceOf[DefaultHttpClient]
```
ed0906: 👍 for cleaning up the GuiceOneAppPerSuite stuff

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -18,24 +18,31 @@ package connectors
...
+
   private val connector = new CustomsDeclareExportsMovementsConnector(config, httpClient)
 
   "Submit Movement" should {
-    given(config.customsDeclareExportsMovements).willReturn(downstreamURL)
 
     "POST to the Back End" in {
       stubFor(
-        post("/movements")
+        post(MovementsSubmissionEndpoint)
```
ed0906: If we change the value of this in the prod code by accident the tests will still pass

I think it was better left as the String as we are asserting the conditions of a contract are met

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -44,39 +51,91 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec {
 
       val request =
         MovementRequest("eori", "provider-id", MovementType.Arrival, ConsignmentReferences("ref", "value"), MovementDetailsRequest("datetime"))
-      await(connector.submit(request))
+      connector.submit(request).futureValue
```
ed0906: Why?

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -44,39 +51,91 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec {
...
 
       verify(
-        postRequestedFor(urlEqualTo("/movements"))
+        postRequestedFor(urlEqualTo(MovementsSubmissionEndpoint))
           .withRequestBody(
             equalTo(
-              "{\"eori\":\"eori\",\"providerId\":\"provider-id\",\"choice\":\"EAL\",\"consignmentReference\":{\"reference\":\"ref\",\"referenceValue\":\"value\"},\"movementDetails\":{\"dateTime\":\"datetime\"}}"
+              """{"eori":"eori","providerId":"provider-id","choice":"EAL","consignmentReference":{"reference":"ref","referenceValue":"value"},"movementDetails":{"dateTime":"datetime"}}"""
```
ed0906: 👍

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -44,39 +51,91 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec {
...
     }
   }
 
   "Submit Consolidation" should {
-    given(config.customsDeclareExportsMovements).willReturn(downstreamURL)
 
     "POST to the Back End" in {
       stubFor(
-        post("/consolidation")
+        post(ConsolidationsSubmissionEndpoint)
```
ed0906: Similar to above

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -44,39 +51,91 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec {
...
           .willReturn(
             aResponse()
               .withStatus(Status.ACCEPTED)
           )
       )
 
       val request = DisassociateDUCRRequest("pid", "eori", "ucr")
-      await(connector.submit(request))
+      connector.submit(request).futureValue
```
ed0906: Why?

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -44,39 +51,91 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec {
...
+          .withRequestBody(equalTo("""{"providerId":"pid","eori":"eori","ucr":"ucr","consolidationType":"DISASSOCIATE_DUCR"}"""))
+      )
+    }
+  }
+
+  "fetch all Submissions" should {
+
+    "send GET request to the backend" in {
+
+      val submissionJson = Json.toJson(Seq(exampleSubmissionFrontendModel()))
```
ed0906: Similar to above, these tests should act as contract tests. So if we are generating the JSON using our own models, the test doesn't really provide much value. It's the reason the above have some JSON strings in the assertions

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -44,39 +51,91 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec {
...
+
+      stubFor(
+        get(s"$FetchAllSubmissionsEndpoint?providerId=$providerId")
+          .willReturn(aResponse().withStatus(OK).withBody(submissionJson.toString))
+      )
+
+      connector.fetchAllSubmissions(providerId).futureValue
```
ed0906: Should we not assert on the response?

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -44,39 +51,91 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec {
...
+    "send GET request to the backend" in {
+
+      val submissionJson = Json.toJson(exampleSubmissionFrontendModel())
+
+      stubFor(
+        get(s"$FetchSingleSubmissionEndpoint/$conversationId?providerId=$providerId")
+          .willReturn(aResponse().withStatus(OK).withBody(submissionJson.toString))
       )
+
+      connector.fetchSingleSubmission(conversationId, providerId).futureValue
```
ed0906: Should we not assert on the response?

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -44,39 +51,91 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec {
...
+    "send GET request to the backend" in {
+
+      val notificationJson = Json.toJson(Seq(exampleNotificationFrontendModel()))
+
+      stubFor(
+        get(s"$FetchNotifications/$conversationId?providerId=$providerId")
+          .willReturn(aResponse().withStatus(OK).withBody(notificationJson.toString))
+      )
+
+      connector.fetchNotifications(conversationId, providerId).futureValue
```
ed0906: Should we not assert on the response?

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -44,39 +51,91 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec {
...
+      val expectedUrl = s"$FetchSingleSubmissionEndpoint/$conversationId?providerId=$providerId"
+      verify(getRequestedFor(urlEqualTo(expectedUrl)))
     }
   }
 
+  "fetch Notifications" should {
+
+    "send GET request to the backend" in {
+
+      val notificationJson = Json.toJson(Seq(exampleNotificationFrontendModel()))
```
ed0906: Similar to above

test/controllers/ViewSubmissionsControllerSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+
+    "call connector for all Notifications" in {
+
+      val submission = MovementsTestData.exampleSubmissionFrontendModel()
+      withConnectorFetchingAllSubmissions(Seq(submission))
+      withConnectorFetchingNotifications(Seq.empty)
+
+      controller.displayPage(getRequest).futureValue
+
+      val expectedProviderId = SuccessfulAuth().operator.pid
```
ed0906: I think `pid` is available as a variable rather than `SuccessfulAuth().operator.pid`

ed0906: If not it should be :)

test/controllers/ViewSubmissionsControllerSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+      val submission3 = exampleSubmissionFrontendModel(requestTimestamp = Instant.now())
+
+      withConnectorFetchingAllSubmissions(Seq(submission1, submission2, submission3))
+      withConnectorFetchingNotifications(Seq.empty)
+
+      controller.displayPage(getRequest).futureValue
+
+      val captor: ArgumentCaptor[Seq[(SubmissionFrontendModel, Seq[NotificationFrontendModel])]] =
+        ArgumentCaptor.forClass(classOf[Seq[(SubmissionFrontendModel, Seq[NotificationFrontendModel])]])
+      verify(submissionsPage).apply(captor.capture())(any(), any())
```
ed0906: Can we extract this into a method like other controller tests do?

It makes the test easier to read

test/testdata/NotificationTestData.scala
```diff
@@ -0,0 +1,47 @@
...
+
+import java.time.Instant
+
+import models.UcrBlock
+import models.notifications.{Entry, NotificationFrontendModel, ResponseType}
+import testdata.CommonTestData.{conversationId, correctUcr}
+
+object NotificationTestData {
+
+  def exampleNotificationFrontendModel(
```
ed0906: `notification(...)`

test/views/SubmissionsViewSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+      page.getElementById("ucr") must containText(messages("submissions.ucr"))
+      page.getElementById("ucrType") must containText(messages("submissions.submissionType"))
+      page.getElementById("submissionAction") must containText(messages("submissions.submissionAction"))
+      page.getElementById("dateOfRequest") must containText(messages("submissions.dateOfRequest"))
+    }
+
+    "contain correct submission data" in {
+
+      val dateTime: Instant = ZonedDateTime
+        .of(LocalDate.parse("2019-10-31", DateTimeFormatter.ofPattern("yyyy-MM-dd")).atStartOfDay(), ZoneId.systemDefault())
```
ed0906: `LocalDate.of(2019,10,31).atStartOfDay().toInstant(ZoneOffset.UTC)`

test/views/SubmissionsViewSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+      page.getElementById(s"ucr-$conversationId").text() mustBe correctUcr
+      page.getElementById(s"ucrType-$conversationId").text() mustBe "DUCR"
+      page.getElementById(s"dateOfRequest-$conversationId").text() mustBe "31 Oct 2019 at 00:00"
+      page.getElementById(s"submissionAction-$conversationId").text() mustBe messages(s"submissions.arrival")
+    }
+
+    "contain MUCR and DUCR if Submission contains both" in {
+
+      val dateTime: Instant = ZonedDateTime
+        .of(LocalDate.parse("2019-10-31", DateTimeFormatter.ofPattern("yyyy-MM-dd")).atStartOfDay(), ZoneId.systemDefault())
```
ed0906: Similar to above

test/views/SubmissionsViewSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+        .toInstant
+      val submission = exampleSubmissionFrontendModel(requestTimestamp = dateTime)
+        .copy(ucrBlocks = Seq(UcrBlock(ucr = correctUcr, ucrType = "M"), UcrBlock(ucr = correctUcr_2, ucrType = "D")))
```
ed0906: Cant we pass in  `ucrBlocks` to the method rather than calling `copy`?

conf/application.conf
```diff
@@ -86,11 +80,6 @@ microservice {
     customs-declare-exports-movements {
       host = localhost
       port = 6797
-      submit-movements = "/movements"
-      submit-consolidation = "/consolidation"
-      fetch-notifications = "/notifications"
-      fetch-all-submissions = "/movements"
-      fetch-single-submission = "/movements"
```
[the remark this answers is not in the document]
ed0906: There's no need for these to exist in config. They're used once and dont change on a per-environment basis

test/connectors/WiremockTestServer.scala
```diff
@@ -41,4 +41,10 @@ trait WiremockTestServer extends WordSpec with BeforeAndAfterAll {
     wireMockServer.stop()
     super.afterAll()
   }
+
+  override def beforeEach(): Unit = {
+    super.beforeEach()
+    wireMockServer.resetAll()
```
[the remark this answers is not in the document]
ed0906: I agree

app/connectors/CustomsDeclareExportsMovementsConnector.scala
MaciejRewera: The whole backend uses `providerId` so I am going to stay with this here.

conf/app.routes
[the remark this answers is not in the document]
MaciejRewera: https://github.com/hmrc/customs-movements-frontend/blob/7c3378b837560301f84d1901c359b14bbfee8314/conf/app.routes#L76

app/connectors/CustomsDeclareExportsMovementsConnector.scala
ed0906: We should be consistent either way and change the `pid`s

conf/app.routes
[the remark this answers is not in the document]
ed0906: This is a new Microservice, we shouldn't copy existing tech debt across if it can be avoided

test/connectors/WiremockTestServer.scala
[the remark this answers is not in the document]
MaciejRewera: But I want to be sure that the mock is "clean" before I start running tests on it, not after.

app/models/notifications/Entry.scala
ed0906: Fair enough, just seems a big of a general name to me, like `Thing` or `Element` etc

MaciejRewera: This is the name given by ILE schema and I'd prefer to stay with the naming even if it's not perfect.

app/models/notifications/NotificationFrontendModel.scala
MaciejRewera: Notification is the model used by backend service to store notifications in the DB. To communicate with frontend it uses different model named `NotificationFrontendModel`. So to be more clear that these models are the same it is called the same on the FE.

app/controllers/ViewSubmissionsController.scala
MaciejRewera: I agree, it would be better. But this would require new endpoint on the BE. So I think this could be done as a separate tech debt ticket.
Also, we need to consider if we are able to handle that huge payload with all notifications in a single call.

ed0906: Fair point I'll raise a ticket

test/connectors/WiremockTestServer.scala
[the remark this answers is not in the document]
ed0906: If you clean it after then your context will always be clean :)

app/connectors/CustomsDeclareExportsMovementsConnector.scala
MaciejRewera: Changed everything to `providerId`

conf/app.routes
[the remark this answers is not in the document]
MaciejRewera: 👍

test/connectors/WiremockTestServer.scala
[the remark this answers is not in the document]
MaciejRewera: I understand your point. In most cases in our services it makes no difference.
My point here is that it is more important to be sure about the state of the mock before the test is run. Or in other words, you should not rely on other tests that they will clean up after themselves.
All in all, this discussion looks like bikeshedding. I think we all might be right, but neither of the solutions is better or worse. Therefore, I am going to stay with my version.

app/models/notifications/NotificationFrontendModel.scala
MaciejRewera: Not in the FE. But for clarity with naming we are using same names as the BE does.

test/base/MockExportsMovementsConnector.scala
MaciejRewera: Agree. This is copy from the external service. Will get rid of it.

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
MaciejRewera: In other services I've only seen `.futureValue`, so consistencyis the first reason. And I've not seen any difference in behavior. Therefore, if there is no functionality differences I'd like to be consistent.

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
MaciejRewera: Answered above.

test/testdata/NotificationTestData.scala
MaciejRewera: As explained earlier, the type returned is `NotificationFrontendModel`, so method name should match it.

test/views/SubmissionsViewSpec.scala
MaciejRewera: No, because this method accepts only ucr and ucrType. This is quite old implementation, before we were aware that there can be more UCRs. However this is a good call to change it now for the new service before it grew too big.

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
MaciejRewera: Yea, that's true.

app/controllers/exchanges/JourneyRequest.scala
```diff
@@ -23,7 +23,7 @@ import play.api.mvc.WrappedRequest
 case class JourneyRequest[T](answers: Answers, request: AuthenticatedRequest[T]) extends WrappedRequest(request) {
 
   val operator: Operator = request.operator
-  val pid: String = request.operator.pid
+  val providerId: String = request.operator.providerId
```
PatrykRudnicki: We can reuse the operator from line above.

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
ed0906: Theres a good split in the codebase of both styles, I don't think it matters really

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -69,14 +74,124 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec {
...
+           |]""".stripMargin
+
+      stubFor(
+        get(s"$FetchAllSubmissionsEndpoint?providerId=$providerId")
+          .willReturn(aResponse().withStatus(OK).withBody(submissionsJson))
+      )
+
+      val response = connector.fetchAllSubmissions(providerId).futureValue
+
+      val expectedUrl = s"$FetchAllSubmissionsEndpoint?providerId=$providerId"
```
ed0906: As mentioned before, think its better just to do 
```
verify(
    getRequestedFor(urlEqualTo("/submissions?providerId=123"))
)
```
That way we test the contract

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -69,14 +74,124 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec {
...
+
+      stubFor(
+        get(s"$FetchSingleSubmissionEndpoint/$conversationId?providerId=$providerId")
+          .willReturn(aResponse().withStatus(OK).withBody(submissionJson))
+      )
+
+      val response = connector.fetchSingleSubmission(conversationId, providerId).futureValue
+
+      val expectedUrl = s"$FetchSingleSubmissionEndpoint/$conversationId?providerId=$providerId"
+      verify(getRequestedFor(urlEqualTo(expectedUrl)))
```
ed0906: As mentioned before, think its better just to do 
```
verify(
    getRequestedFor(urlEqualTo("/submissions/321?providerId=123"))
)
```
That way we test the contract

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -69,14 +74,124 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec {
...
+
+      stubFor(
+        get(s"$FetchNotifications/$conversationId?providerId=$providerId")
+          .willReturn(aResponse().withStatus(OK).withBody(notificationsJson))
+      )
+
+      val response = connector.fetchNotifications(conversationId, providerId).futureValue
+
+      val expectedUrl = s"$FetchNotifications/$conversationId?providerId=$providerId"
+      verify(getRequestedFor(urlEqualTo(expectedUrl)))
```
ed0906: As mentioned before, think its better just to do 
```
verify(
    getRequestedFor(urlEqualTo("/submissions/321?providerId=123"))
)
```
That way we test the contract

test/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
MaciejRewera: Missed those. Fixing.

PR #50, opened 2019-11-11
CEDS-1862 fix not appearing error message on MucrOptions page

MaciejRewera:
While working on fixing this I've noticed that some scenarios are not
handled correctly. Added test suite for MucrOptions.

### DIFF
--- a/app/forms/EnhancedMapping.scala
+++ b/app/forms/EnhancedMapping.scala
@@ -16,8 +16,10 @@
 
 package forms
 
+import forms.MucrOptions.CreateOrAddValues
 import play.api.data.format.Formatter
 import play.api.data.{FieldMapping, FormError, Forms}
+import utils.FieldValidator.validMucr
 
 object EnhancedMapping {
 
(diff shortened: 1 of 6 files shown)


app/forms/EnhancedMapping.scala
```diff
@@ -32,4 +34,52 @@ object EnhancedMapping {
...
     override def unbind(key: String, value: String): Map[String, String] = Map(key -> value)
   }
+
+  object MucrOptionsFormatters {
+
+    lazy val newMucrFormatter: Formatter[String] = new Formatter[String] {
+
+      override def bind(key: String, data: Map[String, String]): Either[Seq[FormError], String] = data.get("createOrAdd") match {
+        case Some(Create) =>
+          data.get("newMucr") match {
```
bambuchaAdm: if you use `key` here you won't have to provide two formatters just only one and use 
```
"newMucr" -> Forms.of(mucrFormatter)
"existingMucr" -> Forms.of(mucrFormatter)
``` 
in mapping

PR #55, opened 2019-11-14
CEDS-1940 Fix Confirmation POST/Redirect/GET logic

ed0906:
It was previously POST -> show view
This means that a refresh of the confirmation screen would re-submit the movement/consolidation

### DIFF
--- a/app/controllers/consolidations/DisassociateUCRController.scala
+++ b/app/controllers/consolidations/DisassociateUCRController.scala
@@ -17,46 +17,44 @@
 package controllers.consolidations
 
 import controllers.actions.{AuthenticatedAction, JourneyRefiner}
-import controllers.storage.FlashKeys
+import forms.DisassociateUcr
 import javax.inject.{Inject, Singleton}
-import models.ReturnToStartException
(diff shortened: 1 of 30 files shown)


app/controllers/consolidations/DisassociateUCRController.scala
```diff
@@ -29,7 +29,7 @@ import views.html.disassociate_ucr
 import scala.concurrent.{ExecutionContext, Future}
 
 @Singleton
-class DisassociateUcrController @Inject()(
+class DisassociateUCRController @Inject()(
```
MaciejRewera: Nit: To me it is more readable when acronyms are not capitalised. You still see that it is an acronym by the word itself and it does not get clumped up with the words that follow.

ed0906: I was just trying to make them all consistent

ed0906: If you want to raise a tech debt thing for the board to discuss how we use Acronyms in class/variable names then go for it :)

MaciejRewera: 👍

PR #56, opened 2019-11-15
CEDS-1940 Acceptance Test  Fixes

ed0906:
Key Changes:
- Ensure ucr type is Upper case e.g. `MUCR` or `DUCR` instead of `mucr` or `ducr`
- Align ViewSubmissions choice code with Movements
- Fix ShutMUCR tab title that just rendered the message key
- `SummaryController` => `MovementSummaryController`

### DIFF
--- a/app/controllers/movements/MovementSummaryController.scala
+++ b/app/controllers/movements/MovementSummaryController.scala
@@ -30,7 +30,7 @@ import views.html.summary.{arrival_summary_page, departure_summary_page}
 
 import scala.concurrent.ExecutionContext
 
-class SummaryController @Inject()(
+class MovementSummaryController @Inject()(
   authenticate: AuthenticatedAction,
   getJourney: JourneyRefiner,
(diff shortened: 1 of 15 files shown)


app/controllers/movements/MovementSummaryController.scala
```diff
@@ -30,7 +30,7 @@ import views.html.summary.{arrival_summary_page, departure_summary_page}
 
 import scala.concurrent.ExecutionContext
 
-class SummaryController @Inject()(
+class MovementSummaryController @Inject()(
```
PatrykRudnicki: Is it worth to change the name to contain `Movement`?

ed0906: huh?

PatrykRudnicki: `SummaryController` IMO was enough. This whole internal service is about movements, so this `Movement` in the name is redundant.

PR #57, opened 2019-11-18
CEDS-1932 Change Transport validation

MaciejRewera:
To give you some context:
Validation on Transport page is different depending on the users choice on GoodsDeparted page. In this PR you can see both validations/mappings added and the logic for providing correct form with the mapping.

There are 2 mappings added for different export scenarios:
1. Out of the UK
2. Back into the UK

Also, there is TransportFormProvider added. To separate logic for
providing correct form and to simplify tests.

### DIFF
--- a/app/forms/Transport.scala
+++ b/app/forms/Transport.scala
@@ -17,12 +17,12 @@
 package forms
 
 import forms.EnhancedMapping.requiredRadio
-import play.api.data.Forms.text
+import play.api.data.Forms.{optional, text}
 import play.api.data.{Form, Forms}
 import play.api.libs.json.Json
 import utils.FieldValidator._
 
-case class Transport(modeOfTransport: String, nationality: String, transportId: String)
+case class Transport(modeOfTransport: Option[String], nationality: Option[String], transportId: Option[String])
 
 object Transport {
   implicit val format = Json.format[Transport]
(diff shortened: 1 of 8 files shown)


app/forms/Transport.scala
```diff
@@ -42,20 +42,45 @@ object Transport {
...
+      "nationality" -> optional(text().verifying("transport.nationality.error", isEmpty or isValidCountryCode)),
+      "transportId" -> optional(
+        text().verifying("transport.transportId.error", isEmpty or (noLongerThan(35) and isAlphanumericWithAllowedSpecialCharacters))
+      )
+    )(Transport.apply)(Transport.unapply)
+    .verifying("transport.backIntoTheUk.error.allFieldsEntered", atLeastOneIsEmpty)
+
+  def form: Form[Transport] = Form(outOfTheUkMapping)
+
+  def outOfTheUkForm: Form[Transport] = Form(outOfTheUkMapping)
```
SteveSugden: Do you need both `form` and `outOfTheUkForm`?

app/forms/providers/TransportFormProvider.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package forms.providers
+
+import forms.GoodsDeparted.DepartureLocation.{BackIntoTheUk, OutOfTheUk}
+import forms.{GoodsDeparted, Transport}
+import models.cache.DepartureAnswers
+import play.api.data.Form
+
+class TransportFormProvider {
```
SteveSugden: Nice!

app/forms/Transport.scala
MaciejRewera: Nope. Good catch! 👍

PR #60, opened 2019-11-21
CEDS-2006/CEDS-2007 Ensure Logging & Errors handled appropriately

ed0906:


app/config/ErrorHandler.scala
```diff
@@ -17,14 +17,28 @@
...
-import play.api.mvc.Request
+import play.api.mvc.{Request, RequestHeader, Result, Results}
 import play.twirl.api.Html
 import uk.gov.hmrc.play.bootstrap.http.FrontendErrorHandler
 import views.html.error
 
 @Singleton
 class ErrorHandler @Inject()(val messagesApi: MessagesApi, errorTemplate: error) extends FrontendErrorHandler {
+
+  private val logger = Logger(classOf[ErrorHandler])
```
PatrykRudnicki: `this.getClass()` will return the same and you will not have a risk that you'll change this class to something different, but this approach is correct too

app/config/ErrorHandler.scala
```diff
@@ -17,14 +17,28 @@
...
+
   override def standardErrorTemplate(pageTitle: String, heading: String, message: String)(implicit request: Request[_]): Html =
     errorTemplate(pageTitle, heading, message)
+
+  override def resolveError(rh: RequestHeader, ex: Throwable): Result = ex match {
```
PatrykRudnicki: I'll put whole name, not shortcuts for `requestHeader` and `exception`

app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -101,6 +95,12 @@ class CustomsDeclareExportsMovementsConnector @Inject()(appConfig: AppConfig, ht
       }
 
   private def providerIdQueryParam(providerId: String): Seq[(String, String)] = Seq("providerId" -> providerId)
+
+  private def logSuccessfulExchange[T](`type`: String, payload: T)(implicit fmt: Format[T]): Unit =
+    logger.debug(`type` + "\n" + Json.toJson(payload))
+
+  private def logFailedExchange(`type`: String, exception: Throwable): Unit =
+    logger.error(`type` + " failed", exception)
```
SteveSugden: Just curious - why error not warn?

MaciejRewera: If I remember correctly, `error` level logs are treated very seriously and might even cause us being called in the middle of the night (?)
I'd rather use `warn` in that case.

ed0906: This happens when the back end returns a non-2xx, in which case we probaly want to know why

ed0906: @MaciejRewera that's only if enough error level logs are raised to reach the PagerDuty threshold

SteveSugden: As long as I don't have to get out of bed, I approve.

bambuchaAdm: On the other hand is this won't be logged on error-handler on top level ?

PR #61, opened 2019-11-21
CEDS-1390 Save and Continue → Continue

bambuchaAdm:
Moveout messages from submit button component

### DIFF
--- a/app/views/components/submit_button.scala.html
+++ b/app/views/components/submit_button.scala.html
@@ -14,7 +14,7 @@
  * limitations under the License.
  *@
 
-@(message: String = "site.save_and_continue")
+@(message: String)
 
 <div class="section">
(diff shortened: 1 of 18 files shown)


app/views/components/submit_button.scala.html
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  *@
 
-@(messageKey: String = "site.save_and_continue")(implicit messages: Messages)
+@(message: String = "site.save_and_continue")
```
PatrykRudnicki: If it's message, the default value doesn't make any sense or should be changed

MaciejRewera: The default value here has not been changed.

bambuchaAdm: Sorry my miss - fixed in next commit

PR #62, opened 2019-11-21
CEDS-PRA Organise Tests into Unit & IT Source Sets

ed0906:


test/util/connectors/ConnectorSpec.scala
```diff
@@ -20,14 +20,12 @@ import base.Injector
...
 import org.scalatest.{BeforeAndAfterEach, MustMatchers, WordSpec}
-import org.scalatestplus.mockito.MockitoSugar
 import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.play.bootstrap.http.DefaultHttpClient
 
 import scala.concurrent.ExecutionContext
 
-class ConnectorSpec
-    extends WordSpec with WiremockTestServer with Injector with MustMatchers with MockitoSugar with BeforeAndAfterEach with ScalaFutures {
+class ConnectorSpec extends WordSpec with WiremockTestServer with Injector with MustMatchers with BeforeAndAfterEach with ScalaFutures {
```
MaciejRewera: Nit: Don't think it requires separate `connector` package if it is already moved to `util`.

test/util/controllers/MessagesStub.scala
```diff
@@ -14,8 +14,9 @@
  * limitations under the License.
  */
 
-package base
+package controllers
```
MaciejRewera: Should not it be `util.controllers` package? Might be the case of IJ moving.

PR #64, opened 2019-11-21
CEDS-PRA Lightweight Choice IT

ed0906:


test/it/ChoiceSpec.scala
```diff
@@ -0,0 +1,48 @@
+import forms.Choice
+import models.cache.{Cache, DepartureAnswers}
+import play.api.http.Status
```
PatrykRudnicki: You don't need to import `Status`. All statuses are in `play.api.tet.Helpers._` which you imported :)

test/it/ChoiceSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+
+      status(response) mustBe Status.FORBIDDEN
+    }
+
+    "return 200" in {
+      givenAuthSuccess()
+
+      val response = get(controllers.routes.ChoiceController.displayPage())
+
+      status(response) mustBe Status.OK
```
PatrykRudnicki: So it'll be just `OK`

PR #68, opened 2019-11-26
CEDS-2009 IT for audit

ed0906:
The principles I've gone with for the Internal UI tests are:

- Send a HTTP Request
- Verify the response code is as expected (ok/redirect)
- Verify the DB is in the expected state
- Verify any requests to other services are done as expected e.g. backend, auth

For the last point, we could consider Auditing as a request to another service. Should we verify these? Or is that going too far?

### DIFF
--- a/app/services/SubmissionService.scala
+++ b/app/services/SubmissionService.scala
@@ -103,7 +103,7 @@ class SubmissionService @Inject()(
     connector.submit(data).map { _ =>
       metrics.incrementCounter(cache.answers.`type`)
       auditService
-        .auditMovements(data, Status.OK.toString, movementAuditType)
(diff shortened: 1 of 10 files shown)


app/services/SubmissionService.scala
```diff
@@ -103,7 +103,7 @@ class SubmissionService @Inject()(
     connector.submit(data).map { _ =>
       metrics.incrementCounter(cache.answers.`type`)
       auditService
-        .auditMovements(data, Status.OK.toString, movementAuditType)
+        .auditMovements(data, "Success", movementAuditType)
```
PatrykRudnicki: It'll be good to extract `Success` as variable

ed0906:
test this please

PR #69, opened 2019-11-27
CEDS-1793 Add visually hidden messages to change buttons

PauloPortugal:


conf/messages.en
```diff
@@ -6,6 +6,10 @@ site.continue = Continue
 site.back = Back
 site.save_and_continue = Save and continue
 site.change = Change
+site.change.hint.disassociate.ucr = Change DUCR dissociation
+site.change.hint.associate.ucr = Change DUCR association
+site.change.hint.associate.mucr = Change MUCR association
+site.change.hint.shut.mucr = Change MUCR
```
MaciejRewera: I would rather use `disassociateUcr` etc. Using dot as a space does not fit in here.

app/views/shut_mucr_summary.scala.html
```diff
@@ -41,7 +41,10 @@ <h2 id="shutMucr-header">@messages("shutMucr.summary.header")</h2>
                     <td id="shutMucr-type">@messages("shutMucr.summary.type")</td>
                     <td id="shutMucr-mucr">@mucr</td>
                     <td class="text-right">
-                        <a id="shutMucr-change" href="@routes.ShutMucrController.displayPage()">@messages("site.edit")</a>
+                        <a id="shutMucr-change" href="@routes.ShutMucrController.displayPage()">
+                            <span aria-hidden="true">@messages("site.edit")</span>
+                            <span class="visuallyhidden">@{messages("site.change.hint.disassociate.ucr")}</span>
```
MaciejRewera: Not sure, but I think it should be `site.change.hint.shut.mucr`

app/views/associate_ucr_summary.scala.html
```diff
@@ -55,7 +58,10 @@ <h2 id="master-consignment">@messages("associate.ucr.summary.masterConsignment")
                     <td class="associate_ucr__to" id="mucr-type">@messages("associate.ucr.summary.kind.mucr")</td>
                     <td id="mucr-reference">@mucr</td>
                     <td class="text-right associate_ucr__change">
-                        <a id="mucr-change" href="@consolidations.routes.MucrOptionsController.displayPage()">@messages("site.change")</a>
+                        <a id="mucr-change" href="@consolidations.routes.MucrOptionsController.displayPage()">
```
MaciejRewera: Could you make id in a similar fashion as the previous one?

conf/messages.en
```diff
@@ -6,6 +6,10 @@ site.continue = Continue
 site.back = Back
 site.save_and_continue = Save and continue
 site.change = Change
+site.change.hint.disassociate.ucr = Change DUCR dissociation
+site.change.hint.associate.ucr = Change DUCR association
+site.change.hint.associate.mucr = Change MUCR association
```
PatrykRudnicki: If we're splitting associate to two different messages (not using one and passing argument. The first one should be probably `ducr` not `ucr`.

```
site.change.hint.associate.ducr = Change DUCR association
site.change.hint.associate.mucr = Change MUCR association
```
or
`site.chage.hint.associate.ucr = Change {0} association`

app/views/shut_mucr_summary.scala.html
PauloPortugal: Correct. Copy paste error.

app/views/associate_ucr_summary.scala.html
PauloPortugal: Done.

app/views/shut_mucr_summary.scala.html
PauloPortugal: Fixed

conf/messages.en
PauloPortugal: I think `disassociate.ucr` reads better than `disassociateUcr`. I have also used the same ID for the `movements` service.

conf/messages.en
PauloPortugal: Good suggestion. Done.

PR #70, opened 2019-11-28
CEDS-1915 retrospective arrivals

MaciejRewera:


app/services/SubmissionService.scala
```diff
@@ -25,15 +25,15 @@ import models.ReturnToStartException
...
+import services.audit.{AuditService, AuditType}
 import uk.gov.hmrc.http.HeaderCarrier
 
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success}
 
 @Singleton
 class SubmissionService @Inject()(
-  cache: CacheRepository,
+  cacheRepository: CacheRepository,
```
ed0906: Every other controller calls this `cache` can we leave it as it is please

test/unit/controllers/movements/ConsignmentReferencesControllerSpec.scala
```diff
@@ -54,7 +54,13 @@ class ConsignmentlReferencesControllerSpec extends ControllerLayerSpec with Mock
   private def theResponseForm: Form[ConsignmentReferences] = {
     val captor = ArgumentCaptor.forClass(classOf[Form[ConsignmentReferences]])
     verify(page).apply(captor.capture())(any(), any())
-    captor.getValue()
+    captor.getValue
+  }
+
+  private def theCache: Cache = {
```
ed0906: You dont need this `MockCache` gives you `theCacheUpserted`

test/unit/controllers/movements/LocationControllerSpec.scala
```diff
@@ -54,7 +54,13 @@ class LocationControllerSpec extends ControllerLayerSpec with MockCache {
   private def theResponseForm: Form[Location] = {
     val captor = ArgumentCaptor.forClass(classOf[Form[Location]])
     verify(page).apply(captor.capture())(any(), any())
-    captor.getValue()
+    captor.getValue
+  }
+
+  private def theCache: Cache = {
```
ed0906: Similar

app/services/SubmissionService.scala
MaciejRewera: I've encountered variable hiding bacause of that, so I'd rather go the other way - to rename it to `cacheRepository` in all controllers.

app/services/audit/AuditService.scala
```diff
@@ -36,19 +37,19 @@ class AuditService @Inject()(connector: AuditConnector, @Named("appName") appNam
 
   def auditShutMucr(providerId: String, mucr: String, result: String)(implicit hc: HeaderCarrier): Future[AuditResult] =
     audit(
-      AuditTypes.AuditShutMucr,
+      AuditType.AuditShutMucr,
       Map(EventData.providerId.toString -> providerId, EventData.mucr.toString -> mucr, EventData.submissionResult.toString -> result)
     )
 
-  def auditDisassociate(providerId: String, ucr: String, result: String)(implicit hc: HeaderCarrier): Future[AuditResult] =
+  def auditDisassociate(providerId: String, ducr: String, result: String)(implicit hc: HeaderCarrier): Future[AuditResult] =
```
bambuchaAdm: are we here able to dissociate only `DUCR`s ?

app/views/components/heading.scala.html
```diff
@@ -23,7 +23,7 @@
   }
 
   @if(title.nonEmpty) {
-    <h1 class="title" id="title">@title</h1>
+    <h1 id="title" class="title">@title</h1>
```
bambuchaAdm: why ?

app/views/movement_confirmation_page.scala.html
```diff
@@ -46,14 +46,14 @@
 
     <h2 id="what-next">@messages("movement.confirmation.whatNext")</h2>
 
-    @if(journeyType == JourneyType.ARRIVE) {
+    @if(Set(JourneyType.ARRIVE, JourneyType.RETROSPECTIVE_ARRIVE).contains(journeyType)) {
         <div id="next-steps">
-            @Html(messages("movement.arrival.confirmation.nextSteps", associateLink, departureLinkForArrival))
+            @Html(s"$associateLink, or $departureLinkForArrival.")
```
bambuchaAdm: Why or ?

test/unit/views/movement/RetrospectiveArrivalSummaryViewSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  "View" should {
+    "render title" in {
+      page(answers).getTitle must containMessage("summary.retrospectiveArrival.title")
+    }
+
+    "render heading" in {
+      page(answers).getElementById("title") must containMessage("summary.retrospectiveArrival.title")
+    }
+
```
bambuchaAdm: test for having sections/inputs on this view ?

app/services/SubmissionService.scala
ed0906: Sounds like the variable should be named differently rather than changing every single controller?

app/services/audit/AuditService.scala
ed0906: Pretty sure this should be `ucr` as it was

app/services/SubmissionService.scala
MaciejRewera: The other variable is of type `Cache` so it is natural to name it `cache`. In this case it is misleading to give variable of type `CacheRepository` the same name. In my opinion it is more correct to name `CacheRepository` instances with something more meaningful.

app/services/audit/AuditService.scala
MaciejRewera: Must have changed that during rebasing. Will change it back.

app/views/components/heading.scala.html
MaciejRewera: Other elements in this component have `id` first, so first to align them. Secondly, I've made a mistake because the order was different for other elements so I thought it would be better to have them all in a simiilar style.

test/unit/controllers/movements/ConsignmentReferencesControllerSpec.scala
MaciejRewera: 👍

test/unit/controllers/movements/LocationControllerSpec.scala
MaciejRewera: 👍

app/views/movement_confirmation_page.scala.html
MaciejRewera: This is a mistake. Fixing it...

app/controllers/ChoiceController.scala
```diff
@@ -33,13 +33,13 @@ import scala.concurrent.{ExecutionContext, Future}
 class ChoiceController @Inject()(
   authenticate: AuthenticatedAction,
   mcc: MessagesControllerComponents,
-  cache: CacheRepository,
+  cacheRepository: CacheRepository,
```
PatrykRudnicki: Not sure that there is any value to use `cacheRepository` than `cache`

PauloPortugal:
LGTM, but it is a long PR to review in detail.

app/services/SubmissionService.scala
```diff
@@ -94,21 +93,30 @@ class SubmissionService @Inject()(
 
   def submit(providerId: String, answers: MovementAnswers)(implicit hc: HeaderCarrier): Future[ConsignmentReferences] = {
     val cache = Cache(providerId, answers)
```
ed0906: There's no need to create a `Cache` here, all you use it for is `cache.answers.type`, you might as well just store the `answers.type` in a variable.

That sorts the need to rename `cache` to `cacheRepository`

app/services/SubmissionService.scala
```diff
@@ -94,21 +93,30 @@ class SubmissionService @Inject()(
...
+    val data = movementBuilder.createMovementExchange(providerId, answers)
     val timer = metrics.startTimer(cache.answers.`type`)
 
-    auditService.auditAllPagesUserInput(answers)
-
-    val movementAuditType =
-      if (cache.answers.`type` == JourneyType.ARRIVE) AuditTypes.AuditArrival else AuditTypes.AuditDeparture
-
-    connector.submit(data).map { _ =>
+    def performSideEffects(result: String)(implicit hc: HeaderCarrier): Unit = {
```
ed0906: performSideEffects??

ed0906: Whatever this is we probably need a more appropriate name & to make it private

app/views/components/heading.scala.html
ed0906: I agree with @MaciejRewera , Doesn't matter in the grand scheme of things but ID should probably go first

app/services/SubmissionService.scala
MaciejRewera: This is an inner method in `submit`. Don't need to be private. And it is an inner method because it requires 4 parameters which are available in the mentioned method.

app/services/SubmissionService.scala
ed0906: What I'm saying here is `submit` & `movementAuditType` can become

```
def submit(providerId: String, answers: MovementAnswers)(implicit hc: HeaderCarrier): Future[ConsignmentReferences] = {
    val data = movementBuilder.createMovementExchange(providerId, answers)
    val timer = metrics.startTimer(answers.`type`)

    def performSideEffects(result: String)(implicit hc: HeaderCarrier): Unit = {
      metrics.incrementCounter(answers.`type`)
      auditService.auditMovements(data, result, movementAuditType(answers.`type`))
      timer.stop()
    }

    auditService.auditAllPagesUserInput(answers)
    connector
      .submit(data)
      .map(_ => cacheRepository.removeByProviderId(providerId))
      .map(_ => data.consignmentReference)
      .andThen {
        case Success(_) => performSideEffects(success)
        case Failure(_) => performSideEffects(failed)
      }

  }

  private def movementAuditType(`type`: JourneyType): AuditType.Value = `type` match {
    case JourneyType.ARRIVE               => AuditType.AuditArrival
    case JourneyType.RETROSPECTIVE_ARRIVE => AuditType.AuditRetrospectiveArrival
    case JourneyType.DEPART               => AuditType.AuditDeparture
  }
```

Then you don't need to create a `Cache`

app/services/SubmissionService.scala
ed0906: Good point it's an inner method. In terms of naming maybe something like:
 `onComplete(success)` & `onComplete(failed)`
or
`on(success) & `on(failed)`

ed0906:
Looks like we need an IT test for this journey

PR #73, opened 2019-12-03
CEDS-1915 retrospective arrivals

MaciejRewera:
Missing update after comments under previous PR.
Also, increased minimum coverage to 80%.

### DIFF
--- a/test/it/IntegrationSpec.scala
+++ b/test/it/IntegrationSpec.scala
@@ -18,7 +18,6 @@ import com.github.tomakehurst.wiremock.client.WireMock
 import com.github.tomakehurst.wiremock.matching.RequestPatternBuilder
 import connectors.{AuditWiremockTestServer, AuthWiremockTestServer, MovementsBackendWiremockTestServer}
 import models.cache.{Answers, Cache}
-import org.apache.bcel.verifier.exc.VerificationException
 import org.scalatest.concurrent.Eventually
 import org.scalatest.{BeforeAndAfterEach, MustMatchers, WordSpec}
 import org.scalatestplus.play.guice.GuiceOneServerPerSuite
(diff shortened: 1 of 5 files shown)


ed0906:
Retro Arrivals still needs an Integration Test :)

MaciejRewera:
That's true. Will add those shortly.

ed0906:
Can we make sure we Squash and Merge this please, the last PR has added a tonne of commits to the git history

MaciejRewera:
DO we realy want to squash these changes? I found it handy to have granular commits when I wanted to revert some changes.

ed0906:
> DO we realy want to squash these changes? I found it handy to have granular commits when I wanted to revert some changes.

We can stick a card on the techtro board to discuss/do a slack vote or something, but generally in the past we've squashed and merged as it keeps the git log clean

MaciejRewera:
Force-pushed after rebase.
The only things added are integration tests for Retrospective Arrivals

test/it/IntegrationSpec.scala
```diff
@@ -70,7 +69,7 @@ abstract class IntegrationSpec
     route(app, request).get
   }
 
-  protected def theCacheFor(pid: String): Option[Answers] = await(cacheRepository.find(Json.obj("providerId" -> "pid")).one[Cache]).map(_.answers)
+  protected def theCacheFor(pid: String): Option[Answers] = await(cacheRepository.find(Json.obj("providerId" -> pid)).one[Cache]).map(_.answers)
```
ed0906: I think this was better as before. Now the `pid` argument isnt used.

Yes it involved us always writing `theCacheFor("pid")` but I like that because it makes it obvious when you glance at the test

ed0906: Sorry ignore me, you fixed it

PR #74, opened 2019-12-03
CEDS-2046 Re-Enable Audit ITs

ed0906:
They were removed previously as they were flakey

This is because the submission service doesn't chain the future returned by the AuditService so it completes in parallel, not necessarily before the verify is called.

The behaviour above is desired, we don't want to couple the success of a submission to whether the Audit Service is up & performant, so I've changed the tests to `verifyEventually` instead of immediately

### DIFF
--- a/test/it/IntegrationSpec.scala
+++ b/test/it/IntegrationSpec.scala
@@ -74,9 +74,6 @@ abstract class IntegrationSpec
 
   protected def givenCacheFor(pid: String, answers: Answers): Unit = await(cacheRepository.insert(Cache.format.writes(Cache(pid, answers))))
 
-  protected def verifyEventually(requestPatternBuilder: RequestPatternBuilder): Unit =
-    eventually {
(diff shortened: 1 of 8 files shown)


test/it/IntegrationSpec.scala
```diff
@@ -67,4 +72,10 @@ abstract class IntegrationSpec
 
   protected def givenCacheFor(pid: String, answers: Answers): Unit = await(cacheRepository.insert(Cache.format.writes(Cache(pid, answers))))
 
+  protected def verifyEventually(requestPatternBuilder: RequestPatternBuilder): Unit = {
+    eventually {
+      Try(WireMock.verify(requestPatternBuilder)).isSuccess mustBe true
+    }
+  }
```
PatrykRudnicki: Brackets for this function are unnecessary, but it's just style comment

ed0906: fixed

PR #75, opened 2019-12-04
CEDS-1915 Fix title on Location page for Retrospective Arrival

MaciejRewera:


PatrykRudnicki:
I prefer messages in lower case not whole word in capital letters, but it's not bad

PR #78, opened 2019-12-05
CEDS-2046 Convert Consignment Reference to an Enum

ed0906:


app/forms/ConsignmentReferences.scala
```diff
@@ -16,59 +16,58 @@
...
+import play.api.libs.json.{Format, Json, Reads, Writes}
 import uk.gov.voa.play.form.ConditionalMappings.mandatoryIfEqual
 import utils.FieldValidator._
 
-case class ConsignmentReferences(reference: String, referenceValue: String)
+case class ConsignmentReferences(reference: ConsignmentReferenceType, referenceValue: String)
+
+object ConsignmentReferenceType extends Enumeration {
+  type ConsignmentReferenceType = Value
+  val D, M = Value
```
bambuchaAdm: Lets stick with Ducr and Mucr with value - makes code less challenging to read

app/utils/FieldValidator.scala
```diff
@@ -87,8 +87,10 @@ object FieldValidator {
 
   val startsWithCapitalLetter: String => Boolean = (input: String) => input.headOption.exists(_.isUpper)
 
-  val isContainedIn: Iterable[String] => String => Boolean =
-    (iterable: Iterable[String]) => (input: String) => iterable.exists(_ == input)
+  def isContainedIn(iterable: Iterable[String]): String => Boolean = (input: String) => iterable.exists(_ == input)
+
+  def isContainedIn[T](iterable: Iterable[T], stringify: T => String): String => Boolean =
+    (input: String) => iterable.map(stringify).exists(_ == input)
```
bambuchaAdm: applying `stringify` on exist allow to kick short circuit on exist - could not copy whole collection if element would be in beginning of collection

app/forms/ConsignmentReferences.scala
bambuchaAdm: Specially if those would be imported into scope

ed0906: It'll involve a bunch of custom Reads & Writes for only minimal gain. Perhaps it can be one for the tech debt board. It's still better than what we have today :)

ed0906: I'll add a formatter test though so it's easier to do

app/utils/FieldValidator.scala
ed0906: I dont think we need to be optimizing to the nanosecond ;)

PR #81, opened 2019-12-12
BAU - refactored Title to ensure that translated messages are not used

SteveSugden:
Title translates all parts of the title and as all view tests extend
ViewSpec (which uses 'AllMessageKeysAreMandatoryMessages') then the
message keys are validated to exist in the messages file.


test/unit/views/movement/ConsignmentReferenceViewSpec.scala
```diff
@@ -31,7 +31,7 @@ class ConsignmentReferenceViewSpec extends ViewSpec {
 
     "render title" in {
 
-      page(ConsignmentReferences.form()).getTitle must containMessage("consignmentReferences.title")
+      page(ConsignmentReferences.form()).getTitle must containMessage("consignmentReferences.ARRIVE.question")
```
PatrykRudnicki: I know it's not a part of this PR but I would prefer to have `arrive` not `ARRIVE`

SteveSugden: Normally I would point out that we have used enum raw values in a few places as part of the message keys.  But as we have time atm I'll change it.   
Just don't expect this level of service all of the time.

app/views/Title.scala
```diff
@@ -18,13 +18,16 @@ package views
 
 import play.api.i18n.Messages
 
-case class Title(title: String)
+case class Title(headingKey: String, sectionKey: Option[String] = None, headingArgs: Seq[String] = Seq.empty) {
```
bambuchaAdm: What do you think about use empty string as null object? code would be simple - less pointless `Some` in code

SteveSugden: I guess so.  Defaulted to "".

PR #84, opened 2019-12-30
CEDS Update tempermoneky to work correctly on all pages

bambuchaAdm:


bambuchaAdm:
From programing perspective i think that would too much logic in lower layer of script. I agree there is missing docs that `isCurrentPage` take regexp but not string.

PatrykRudnicki:
Old PR which has conflicts.

Probably this functionality is covered by the actual QS script

PR #87, opened 2020-01-13
CEDS-2091 Connector for ILE endpoint on backend

PatrykRudnicki:


app/connectors/exchanges/IleQueryRequest.scala
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package connectors.exchanges
+
+import models.UcrBlock
+import play.api.libs.json.Json
+
+case class IleQueryRequest(eori: String, providerId: Option[String], ucrBlock: UcrBlock)
```
MaciejRewera: All other Exchange classes has `providerId` as just `String`. Is there a particular reason that it is optional here?
And one more thing: for consistency, maybe it would be better to call this class `IleQueryExchange`?

test/unit/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
```diff
@@ -204,4 +205,29 @@ class CustomsDeclareExportsMovementsConnectorSpec extends ConnectorSpec with Moc
...
+      val result = connector.submit(request).futureValue
+
+      verify(
+        postRequestedFor(urlEqualTo("/consignment-query"))
+          .withRequestBody(equalToJson("""{"eori":"eori","providerId":"provider-id","ucrBlock":{"ucr":"ucr","ucrType":"D"}}""".stripMargin))
+      )
+
+      result mustBe conversationId
+    }
+  }
```
MaciejRewera: Nit: Could you move this test to be next to other submission tests?

app/connectors/exchanges/IleQueryRequest.scala
PatrykRudnicki: I used this to be aligned with the frontend. Good point. I changed to be just a String and changed name.

test/unit/connectors/CustomsDeclareExportsMovementsConnectorSpec.scala
PatrykRudnicki: Fixed

PR #89, opened 2020-01-15
CEDS-2091 Ile query implementation

PatrykRudnicki:


app/controllers/ileQuery/IleQueryController.scala
```diff
@@ -0,0 +1,81 @@
...
+        }
+      )
+
+  private def buildIleQuery(providerId: String, ucr: String): IleQueryExchange = {
+    val ucrType = if (validDucr(ucr)) "D" else "M"
+
+    val ucrBlock = UcrBlock(ucr, ucrType)
+
+    IleQueryExchange(Answers.fakeEORI.get, providerId, ucrBlock)
+  }
```
MaciejRewera: I would move this method away to a separate class. It requires to be deeply tested, especially the first line of the method. It will be a headache and a bad practise to test it through controller.

app/controllers/ileQuery/IleQueryController.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  def submit(): Action[AnyContent] = authenticate { implicit request =>
+    IleQuery.form.fold(
+      formWithErrors => BadRequest(ileQueryPage(formWithErrors)),
+      validUcr => Redirect(controllers.ileQuery.routes.IleQueryController.submitQuery(validUcr))
+    )
+  }
+
+  def submitQuery(ucr: String): Action[AnyContent] = authenticate.async { implicit request =>
```
MaciejRewera: I am still thinking that this method should be either in a separate controller, or to rename `submit` or `submitQuery` to clearly show what they are doing and which part of the journey they conver.
Currently their names could be swapped with no implications.

app/controllers/ileQuery/IleQueryController.scala
```diff
@@ -0,0 +1,81 @@
...
+    request.cookies.get("conversationId") match {
+      case Some(cookie) =>
+        connector.fetchQueryNotification(cookie.value, request.providerId).map { notifications =>
+          if(notifications.isEmpty) Ok("loading").withHeaders("refresh" -> "1")
```
MaciejRewera: It would be good to be able to controll the refresh rate. At least through the config.

app/controllers/ileQuery/IleQueryController.scala
```diff
@@ -0,0 +1,81 @@
...
+    IleQuery.form.fold(
+      formWithErrors => BadRequest(ileQueryPage(formWithErrors)),
+      validUcr => Redirect(controllers.ileQuery.routes.IleQueryController.submitQuery(validUcr))
+    )
+  }
+
+  def submitQuery(ucr: String): Action[AnyContent] = authenticate.async { implicit request =>
+    request.cookies.get("conversationId") match {
+      case Some(cookie) =>
+        connector.fetchQueryNotification(cookie.value, request.providerId).map { notifications =>
```
MaciejRewera: Nit: To have visual balance, I would move the part when there is a cookie to a private method.

app/models/cache/JourneyType.scala
```diff
@@ -21,7 +21,7 @@ import play.api.libs.json.{Format, Reads, Writes}
 object JourneyType extends Enumeration {
   type JourneyType = Value
 
-  val ARRIVE, RETROSPECTIVE_ARRIVE, DEPART, ASSOCIATE_UCR, DISSOCIATE_UCR, SHUT_MUCR = Value
+  val ARRIVE, RETROSPECTIVE_ARRIVE, DEPART, ASSOCIATE_UCR, DISSOCIATE_UCR, SHUT_MUCR, ILE_QUERY = Value
```
MaciejRewera: What is that required for?

PatrykRudnicki:
@MaciejRewera Idea with IleQuery service good. I'll extract some parts to service. I don't see value in separate controllers, specially that methods around form are really small. I'll try to change the naming to have better description

app/controllers/ileQuery/IleQueryController.scala
PatrykRudnicki: Based on the BAs discussion I think this method is strictly connected with our form. I know that the better place is to have it in separate service, but not sure that this is the best solution in that specific case

app/controllers/ileQuery/IleQueryController.scala
PatrykRudnicki: The first two methods are really small, I don't see value to extracting this to the new controller but I can do it.

I'll change the names to be more describing.

app/controllers/ileQuery/IleQueryController.scala
PatrykRudnicki: Yup, I'll do it

app/models/cache/JourneyType.scala
PatrykRudnicki: You're right. I made a mistake, it should be `ActionType`

app/config/AppConfig.scala
```diff
@@ -48,4 +50,6 @@ class AppConfig @Inject()(
 
   lazy val customsDeclarationsGoodsTakenOutOfEuUrl: String = loadConfig("urls.customsDeclarationsGoodsTakenOutOfEu")
   lazy val serviceAvailabilityUrl: String = loadConfig("urls.serviceAvailability")
+
+  lazy val cacheTimeToLive: FiniteDuration = servicesConfig.getDuration("mongodb.timeToLive").asInstanceOf[FiniteDuration]
```
bambuchaAdm: it is more ile-query related option rather than mongodb

app/controllers/ileQuery/IleQueryController.scala
```diff
@@ -0,0 +1,125 @@
...
+      .fold(
+        formWithErrors => Future.successful(BadRequest(ileQueryPage(formWithErrors))),
+        validUcr => {
+          val ileQueryRequest = buildIleQuery(request.providerId, validUcr)
+
+          connector.submit(ileQueryRequest).flatMap { conversationId =>
+            val ileQuery = IleQuery(retrieveSessionId, validUcr, conversationId)
+
+            ileQueryRepository.insert(ileQuery).map { _ =>
+              Redirect(controllers.ileQuery.routes.IleQueryController.submitQuery(ucr))
```
bambuchaAdm: why not loading screen with refresh - we would make here another request

app/forms/IleQuery.scala
```diff
@@ -0,0 +1,32 @@
...
+
+package forms
+
+import play.api.data.{Form, Forms}
+import play.api.data.Forms.text
+import utils.FieldValidator._
+
+object IleQuery {
+
+  def form: Form[String] = Form(
```
bambuchaAdm: val ?

app/models/cache/IleQuery.scala
```diff
@@ -0,0 +1,42 @@
...
+object IleQuery {
+
+  implicit val formatInstant: OFormat[Instant] = new OFormat[Instant] {
+    override def writes(datetime: Instant): JsObject =
+      Json.obj("$date" -> datetime.toEpochMilli)
+
+    override def reads(json: JsValue): JsResult[Instant] =
+      json match {
+        case JsObject(map) if map.contains("$date") =>
+          map("$date") match {
```
bambuchaAdm: We can do this in single map lookup i think
```
map.get("$date") match {
  case Some(JsNumber(v)) =>
  case Some(_) =>
  case None =>
}
```

app/controllers/ileQuery/IleQueryController.scala
```diff
@@ -0,0 +1,125 @@
...
+    ileQueryRepository.findBySessionIdAndUcr(retrieveSessionId, ucr).flatMap {
+      case Some(query) =>
+        connector.fetchQueryNotifications(query.conversationId, request.providerId).flatMap { response =>
+          response.status match {
+            case OK =>
+              ileQueryRepository.removeByConversationId(query.conversationId).map { _ =>
+                val queryResponse = Json.parse(response.body).as[IleQueryResponse]
+
+                val ducrResult = queryResponse.queriedDucr.map(ducr => Ok(ileQueryDucrResponsePage(ducr)))
+                val mucrResult = queryResponse.queriedMucr.map(mucr => Ok(ileQueryMucrResponsePage(mucr)))
```
SteveSugden: I particularly like this bit

app/views/loading_screen.scala.html
```diff
@@ -0,0 +1,31 @@
...
+@()(implicit request: Request[_], messages: Messages)
+
+@main_template(title = Title("ileQuery.loading.title")) {
+
+    @components.page_title(
+        Some(messages("ileQuery.loading.title")),
+        Some(messages("ileQuery.loading.hint"))
+    )
+
+    Spinner
```
SteveSugden: Is "Spinner" a TODO?

conf/testOnlyDoNotUseInAppConf.routes
```diff
@@ -11,3 +11,9 @@
 
 # Add all the application routes to the prod.routes file
 ->         /                          prod.Routes
+
+# ILE Query
+GET         /customs-exports-internal/consignment-query                     controllers.ileQuery.IleQueryController.displayQueryForm()
```
bambuchaAdm: Why duplicated ?

test/util/base/UnitSpec.scala
```diff
@@ -18,5 +18,9 @@ package base
 
 import org.scalatest.{MustMatchers, WordSpec}
 import org.scalatestplus.mockito.MockitoSugar
+import reactivemongo.api.commands.{LastError, WriteResult}
 
-trait UnitSpec extends WordSpec with MustMatchers with MockitoSugar
+trait UnitSpec extends WordSpec with MustMatchers with MockitoSugar {
+  val dummyWriteResultSuccess: WriteResult =
```
bambuchaAdm: Why in UnitSpec ??

app/controllers/ileQuery/IleQueryController.scala
SteveSugden: I'm still confused by the method name.  Only when I got to the routes file did I realised that this is a "GET" endpoint.

Maybe this should be `displayQueryResponse' ?   (I realise that the outcome of calling this end-point is not always the query response - but that is the main intention.

conf/application.conf
```diff
@@ -15,7 +15,7 @@
 include "frontend.conf"
 
 appName = "customs-exports-internal-frontend"
-application.router = prod.Routes
+application.router = testOnlyDoNotUseInAppConf.Routes
```
SteveSugden: Is this effectively the feature switch that prevents these new endpoints being exposed?   Do we need to update one of the other app-confix-xxx projects?

app/config/AppConfig.scala
MaciejRewera: Agree with the above

app/models/cache/IleQuery.scala
```diff
@@ -0,0 +1,42 @@
...
+
+package models.cache
+
+import java.time.{Instant, ZoneOffset}
+
+import play.api.libs.json.{JsError, JsNumber, JsObject, JsResult, JsSuccess, JsValue, Json, OFormat}
+
+case class IleQuery(sessionId: String, ucr: String, conversationId: String, createdAt: Instant = Instant.now())
+
+object IleQuery {
```
MaciejRewera: This has the same name as `forms.IleQuery`. I'd change one of them.

app/models/cache/IleQuery.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  implicit val formatInstant: OFormat[Instant] = new OFormat[Instant] {
```
MaciejRewera: Could this possibly be `private` ?

app/config/AppConfig.scala
PatrykRudnicki: Changed

app/controllers/ileQuery/IleQueryController.scala
PatrykRudnicki: Because we can have a really fast response. This will allow to first check notifications before showing loading screen which can be faster

PatrykRudnicki: Above approach discussed with BAs and agreed that it's better to check if there is some notifications before showing loading screen

app/forms/IleQuery.scala
PatrykRudnicki: Changed

app/models/cache/IleQuery.scala
PatrykRudnicki: In that case we will miss UnexpectedDateFormat. Not sure that's necessary.

app/views/loading_screen.scala.html
PatrykRudnicki: Yes, Thank you!

conf/testOnlyDoNotUseInAppConf.routes
PatrykRudnicki: Why duplicated? What do you mean?

test/util/base/UnitSpec.scala
PatrykRudnicki: We're using this in few places and every place extends `UnitSpec`. This is probably the only one existing place where we can put it. Different approach will be new trait and extending this trait in specific tests or creating this every time in specific test

conf/application.conf
PatrykRudnicki: Yes, we will allow this on Development, but after merge

app/models/cache/IleQuery.scala
PatrykRudnicki: yes, agree. What propose? IleQueryForm?

app/models/cache/IleQuery.scala
PatrykRudnicki: Probably, but I don't see any reason in doing it. All formats in our companion objects are public.

app/views/ViewDates.scala
```diff
@@ -0,0 +1,27 @@
...
+
+package views
+
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
+
+object ViewDates {
+  def timezone = ZoneId.of("Europe/London")
+
+  val movementFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern("dd/MM/yyyy").withZone(timezone)
```
MaciejRewera: So far, dateTime formattingwas handled by injected DateTimeFormatter.
I'd advise to use same approach here. Bear in mind however, that there is one `DateTimeModule` already

PatrykRudnicki: @SteveSugden Can you look at this comment? I don't know why we have this

app/views/IleCodeMapper.scala
```diff
@@ -0,0 +1,45 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package views
+import play.api.i18n.Messages
+
+object IleCodeMapper {
```
MaciejRewera: There are code classes in `models.viewmodels.decoder` package for ROE, SOE and ICS codes. But as I can see you cannot use them straight away, as the codes are translated to different messages

app/controllers/ileQuery/IleQueryController.scala
```diff
@@ -0,0 +1,125 @@
...
+
+                ducrResult.orElse(mucrResult).getOrElse(loadingPageResult)
+              }
+            case NO_CONTENT =>
+              Future.successful(loadingPageResult)
+            case GATEWAY_TIMEOUT =>
+              ileQueryRepository.removeByConversationId(query.conversationId).map { _ =>
+                InternalServerError(errorHandler.standardErrorTemplate())
+              }
+            case _ => Future.successful(BadRequest(errorHandler.standardErrorTemplate()))
```
MaciejRewera: I don't think this a case for `BadRequest`. To me it is more like `InternalServerError`. So maybe just throw an exception in this case? Or "let it crash" by having `MatchException`

conf/testOnlyDoNotUseInAppConf.routes
bambuchaAdm: you have same routes in prod routes

conf/application.conf
bambuchaAdm: This should be made only on app-config-development

app/views/ViewDates.scala
SteveSugden: Yes.  It took me ages to work out where the DateTimeFormatter was magically coming from in other forms.   Surely it is better for the view to specify what formater it wants to use.  This helper class is just a place to define a format so that it can be re-used by other views if required.

app/controllers/ileQuery/IleQueryController.scala
PatrykRudnicki: Will change to `InternalServerError`

app/models/cache/IleQuery.scala
bambuchaAdm: But this format is quite local to this - do we expect that this format should be used somewhere else? If not i don't see why we should not go into private.

app/views/ViewDates.scala
bambuchaAdm: On the other hand - Dependency Injection is usual way to obtain those type of objects.

SteveSugden: Isn't DI more useful for injecting complex components that would make the consumer component harder to test otherwise?

I think this is just adding complexity for no apparent gain.  However...

Happy to change it if someone can explain how we over-ride the formatter returned by DateTimeModule.    Do we have an example of this.

app/controllers/ileQuery/IleQueryController.scala
```diff
@@ -0,0 +1,140 @@
...
+                InternalServerError(errorHandler.standardErrorTemplate())
+              }
+          }
+        }
+
+      case None => sendIleQuery(ucr)
+    }
+  }
+
+  private def loadingPageResult()(implicit request: Request[AnyContent]) =
```
bambuchaAdm: is this code is use somewhere else then submitQuery ? I think local def's are better then private methods if there single usage of this code.

PatrykRudnicki: Yes, I agree with you, but now it's used in two different methods. Still see the pros of extracting this in one place to have the refresh value to change in one place.

I'll try to refactor this page after this PR

conf/testOnlyDoNotUseInAppConf.routes
PatrykRudnicki: Fixed

app/views/loading_screen.scala.html
PatrykRudnicki: Spinner added

app/controllers/ileQuery/IleQueryController.scala
PatrykRudnicki: Changed

app/models/cache/IleQuery.scala
PatrykRudnicki: Changed to private

app/models/cache/IleQuery.scala
PatrykRudnicki: Changed Form to have different name

PR #93, opened 2020-01-27
CEDS-2092 - View consignment details for a DUCR or MUCR

SteveSugden:


app/controllers/ileQuery/IleQueryController.scala
```diff
@@ -67,9 +71,15 @@ class IleQueryController @Inject()(
           response.status match {
             case OK =>
               ileQueryRepository.removeByConversationId(query.conversationId).map { _ =>
-                val notifications = Json.parse(response.body).validate[Seq[NotificationFrontendModel]]
-
-                Ok("Result" + notifications)
+                val queryResponse = Json.parse(response.body).validate[IleQueryResponse].get
```
bambuchaAdm: `validate[t].get` is equivalent of `as`

app/controllers/ileQuery/IleQueryController.scala
```diff
@@ -67,9 +71,15 @@ class IleQueryController @Inject()(
...
+
+                if (queryResponse.queriedDucr.isDefined) {
```
bambuchaAdm: Maybe `map` + `orElse` ?

PR #94, opened 2020-01-27
CEDS-2098 - Associate UCR page

iservport:
This is a feature branch to consolidate merges from all related page-specific branches:

* CEDS-2098-arrive (affects departure and retrospective arrival) - ok
* CEDS-2098-associate - ok
* CEDS-2098-disassociate - TODO
* CEDS-2098-shut - TODO
* CEDS-2098-clean - TODO

It is an alternative confirmation screen to implement in the internal service as a way of avoiding cases being closed when they shouldn't be.

Currently, NCH operators see the existing confirmation screen at the end of each movement request journey and believe that it means their request has been completed and accepted. As their request has only been submitted to DMS and the acceptance (or rejection) is asynchronous, this could be misleading.

### DIFF
--- a/app/views/associate_ucr_confirmation.scala.html
+++ b/app/views/associate_ucr_confirmation.scala.html
@@ -14,33 +14,33 @@
  * limitations under the License.
  *@
 
-@import controllers.routes
-@import uk.gov.hmrc.play.views.html._
-@import controllers.storage.FlashKeys
 @import views.Title
-@import views.html.templates.main_template
+@import components.gds.gds_main_template
+@import components.gds.pageTitle
+@import uk.gov.hmrc.govukfrontend.views.html.components.govukInsetText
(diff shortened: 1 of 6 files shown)


app/views/associate_ucr_confirmation.scala.html
```diff
@@ -14,33 +14,33 @@
...
-
-    <div id="next-steps">
-        @Html(messages("association.confirmation.associateOrDepart", associateLink, shutMucrLink))
-    </div>
-
-    @components.button_link("site.backToStart", routes.ChoiceController.displayPage())
-}
\ No newline at end of file
+    <p class="govuk-body">
+    @link(message = messages("movement.confirmation.redirect.link"), href = routes.ChoiceController.displayPage())
```
SteveSugden: Does this message key exist in feature branch from previous PR?

iservport: Yes

PR #96, opened 2020-01-28
CEDS-2098 - Shut MUCR page

iservport:
This is a feature branch to consolidate merges from all related page-specific branches:

* CEDS-2098-arrive (affects departure and retrospective arrival) - ok
* CEDS-2098-associate - ok
* CEDS-2098-disassociate - ok
* CEDS-2098-shut - ok
* CEDS-2098-clean - TODO

It is an alternative confirmation screen to implement in the internal service as a way of avoiding cases being closed when they shouldn't be.

Currently, NCH operators see the existing confirmation screen at the end of each movement request journey and believe that it means their request has been completed and accepted. As their request has only been submitted to DMS and the acceptance (or rejection) is asynchronous, this could be misleading.

### DIFF
--- a/test/unit/controllers/consolidations/ShutMUCRConfirmationControllerSpec.scala
+++ b/test/unit/controllers/consolidations/ShutMUCRConfirmationControllerSpec.scala
@@ -19,17 +19,20 @@ package controllers.consolidations
 import controllers.ControllerLayerSpec
 import controllers.actions.AuthenticatedAction
 import controllers.storage.FlashKeys
-import models.ReturnToStartException
+import org.mockito.ArgumentMatchers.any
+import org.mockito.Mockito.when
+import org.scalatestplus.mockito.MockitoSugar
 import play.api.http.Status
 import play.api.test.FakeRequest
 import play.api.test.Helpers._
+import play.twirl.api.HtmlFormat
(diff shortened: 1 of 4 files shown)


test/unit/controllers/consolidations/ShutMUCRConfirmationControllerSpec.scala
```diff
@@ -19,17 +19,20 @@ package controllers.consolidations
...
+import play.twirl.api.HtmlFormat
 import views.html.shut_mucr_confirmation
 
 import scala.concurrent.ExecutionContext.Implicits.global
 
-class ShutMUCRConfirmationControllerSpec extends ControllerLayerSpec {
+class ShutMUCRConfirmationControllerSpec extends ControllerLayerSpec with MockitoSugar {
 
-  private val page = new shut_mucr_confirmation(main_template)
+  private val page = mock[shut_mucr_confirmation]
```
PatrykRudnicki: I know that you're using this mock here and only on one test case, but in case that we will add more tests is better to use `beforeEach()` to mock behaviour of the page and `afterEach()` to reset the page.

iservport: Agreed!

PR #99, opened 2020-01-29
CEDS-2102 - store queried ucr and type as preparation for journeys

SteveSugden:
This PR saves the queried ucr and type as part of the viewing of ILE Query Results to the users cache.  
Nothing is done with that saved data at present, but it will be used in future to integrate the queried ucr into the various journeys (i.e. to pre-fill questions or even skip pages)

### DIFF
--- a/app/controllers/actions/JourneyRefiner.scala
+++ b/app/controllers/actions/JourneyRefiner.scala
@@ -35,7 +35,7 @@ class JourneyRefiner @Inject()(cacheRepository: CacheRepository)(implicit val ex
     cacheRepository.findByProviderId(request.operator.providerId).map(_.flatMap(_.answers)).map {
       case Some(answers: Answers) if types.isEmpty || types.contains(answers.`type`) =>
         Right(JourneyRequest(answers, request))
-      case a =>
+      case _ =>
         Left(Results.Redirect(controllers.routes.ChoiceController.displayPage()))
     }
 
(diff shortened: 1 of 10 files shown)


app/controllers/actions/JourneyRefiner.scala
```diff
@@ -32,10 +32,11 @@ class JourneyRefiner @Inject()(cacheRepository: CacheRepository)(implicit val ex
   override protected def executionContext: ExecutionContext = exc
 
   private def refiner[A](request: AuthenticatedRequest[A], types: JourneyType*): Future[Either[Result, JourneyRequest[A]]] =
-    cacheRepository.findByProviderId(request.operator.providerId).map(_.map(_.answers)).map {
+    cacheRepository.findByProviderId(request.operator.providerId).map(_.flatMap(_.answers)).map {
       case Some(answers: Answers) if types.isEmpty || types.contains(answers.`type`) =>
         Right(JourneyRequest(answers, request))
-      case _ =>
+      case a =>
+        print(s"foud this " + a)
```
PatrykRudnicki: If we want to merge this PR please remove this case

SteveSugden: Oops!

app/controllers/actions/JourneyRefiner.scala
```diff
@@ -36,7 +36,6 @@ class JourneyRefiner @Inject()(cacheRepository: CacheRepository)(implicit val ex
       case Some(answers: Answers) if types.isEmpty || types.contains(answers.`type`) =>
         Right(JourneyRequest(answers, request))
       case a =>
```
PatrykRudnicki: `case a` is necessary? ``Probably `case _ => ` will be better

PatrykRudnicki: I mean previous implementation :)

SteveSugden: Sorry - I just saw the print statement.   Fixed now

PR #101, opened 2020-01-30
CEDS 2102 - pre-populate query ucr in journeys

SteveSugden:
This draft PR completes the CEDS-2102 spike - pre-populating the queried ucr (ducr/mucr) into the various journeys.

If we are happy with this approach and can complete with tests.

### DIFF
--- a/app/models/UcrBlock.scala
+++ b/app/models/UcrBlock.scala
@@ -22,4 +22,7 @@ case class UcrBlock(ucr: String, ucrType: String)
 
 object UcrBlock {
   implicit val format = Json.format[UcrBlock]
+
+  val ducr = "D"
+  val mucr = "M"
 }
--- a/app/models/cache/Answers.scala
+++ b/app/models/cache/Answers.scala
@@ -17,6 +17,8 @@
 package models.cache
 
 import forms.{AssociateUcr, MucrOptions, _}
+import models.UcrBlock
+import models.UcrBlock.{ducr, mucr}
 import models.cache.JourneyType.JourneyType
 import play.api.libs.json.{Format, Json}
 import uk.gov.hmrc.play.json.Union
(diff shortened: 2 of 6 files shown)


app/models/UcrBlock.scala
```diff
@@ -22,4 +22,7 @@ case class UcrBlock(ucr: String, ucrType: String)
 
 object UcrBlock {
   implicit val format = Json.format[UcrBlock]
+
+  val ducr = "D"
+  val mucr = "M"
```
MaciejRewera: Maybe better call them `DucrType` and `MucrType` ?

SteveSugden: Can do.  And I don't think I've used them consistently - there are still some hard-coded "D" and "M" out there.

Doing this exercise highlighted how many times we have created constants/objects/enums to represent ducr or mucr in the product though!  :-)

app/models/cache/Answers.scala
```diff
@@ -57,6 +59,15 @@ case class ArrivalAnswers(
 
 object ArrivalAnswers {
   implicit val format: Format[ArrivalAnswers] = Json.format[ArrivalAnswers]
+
+  def apply(queryUcr: Option[UcrBlock]): ArrivalAnswers = queryUcr match {
+    case Some(ucrBlock) if ucrBlock.ucrType.equals(mucr) =>
+      ArrivalAnswers(consignmentReferences = Some(ConsignmentReferences(ConsignmentReferenceType.M, ucrBlock.ucr)))
```
MaciejRewera: Could you please use parameter names for `ConsignmentReferences` ?

SteveSugden: You mean `ConsignmentReferences(reference = ConsignmentReferenceType.M, referenceValue = ucrBlock.ucr))` ?  Is that necessary for clarity?

PatrykRudnicki: I think it's not necessary in cases where we're filling all arguments and additionally those arguments are really good described. I mean `ConsignmentReferenceType.M` `ucrBlock.ucr`

PatrykRudnicki: Why you're not using `UcrBlock.ducr`, but `ConsignmentReferenceType.D` ?

SteveSugden: Don't understand . 
UcrBlock.mucr => ConsignmentReferenceType.M
UcrBlock.ducr => ConsignmentReferenceType.D

SteveSugden:
Thanks for the comments.  I've demo'd to Ryan and he's happy to proceed so I'll work this up into a proper PR with tests.

app/models/cache/Answers.scala
PatrykRudnicki: My fault, misunderstood. It's okay. I thought that UcrBlock.mucr has the same value like ConsignmentReferenceType.M

PR #102, opened 2020-01-30
CEDS-2102 Pre-populate the UCR in movements journeys

SteveSugden:
Takes the optional UcrBlock resulting from an ILE query response and uses the ducr or mucr to pre-populate the relevant ucr field in the initial "Answers" for the users selected journey.

### DIFF
--- a/app/controllers/ChoiceController.scala
+++ b/app/controllers/ChoiceController.scala
@@ -41,8 +41,10 @@ class ChoiceController @Inject()(
   def displayPage: Action[AnyContent] = authenticate.async { implicit request =>
     cacheRepository.findByProviderId(request.providerId).map {
       case Some(cache) =>
-        cache.answers.map(answers => Ok(choicePage(Choice.form().fill(Choice(answers.`type`))))).getOrElse(Ok(choicePage(Choice.form())))
-      case None => Ok(choicePage(Choice.form()))
+        cache.answers
+          .map(answers => Ok(choicePage(Choice.form().fill(Choice(answers.`type`)), cache.queryUcr)))
+          .getOrElse(Ok(choicePage(Choice.form(), cache.queryUcr)))
+      case None => Ok(choicePage(Choice.form(), None))
     }
   }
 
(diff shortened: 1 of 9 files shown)


SteveSugden:
No feature flag required for this.    This functionality is driven by the presence of data in the users cache that is only added as a result of executing an ILE query and that is protected by the prod.routes.

app/models/UcrBlock.scala
```diff
@@ -22,4 +22,7 @@ case class UcrBlock(ucr: String, ucrType: String)
 
 object UcrBlock {
   implicit val format = Json.format[UcrBlock]
+
+  val ducrType = "D"
+  val mucrType = "M"
```
MaciejRewera: 👍

app/controllers/ChoiceController.scala
```diff
@@ -60,15 +61,31 @@ class ChoiceController @Inject()(
...
-    case Choice.AssociateUCR         => saveAndRedirect(AssociateUcrAnswers(), consolidations.routes.MucrOptionsController.displayPage())
-    case Choice.DisassociateUCR      => saveAndRedirect(DisassociateUcrAnswers(), consolidations.routes.DisassociateUCRController.display())
-    case Choice.ShutMUCR             => saveAndRedirect(ShutMucrAnswers(), consolidations.routes.ShutMucrController.displayPage())
-    case Choice.ViewSubmissions      => Future.successful(Redirect(routes.ViewSubmissionsController.displayPage()))
+    case Choice.Arrival =>
+      saveAndRedirect(ArrivalAnswers.fromQueryUcr, movements.routes.ConsignmentReferencesController.displayPage())
+    case Choice.RetrospectiveArrival =>
+      saveAndRedirect(RetrospectiveArrivalAnswers.fromQueryUcr, movements.routes.ConsignmentReferencesController.displayPage())
+    case Choice.Departure =>
+      saveAndRedirect(DepartureAnswers.fromQueryUcr, movements.routes.ConsignmentReferencesController.displayPage())
```
MaciejRewera: Nit: I like the idea with using function parameter. To me it would be more natural to add `fromQueryUcr` method to `Answers` trait and then just call it in `saveAndRedirect`. The signature of this method would stay the same in that case.
But I am just more used to OOP solutions.

PR #108, opened 2020-02-03
CEDS-2105 - Create a placeholder page if UCR is not present in ILE

iservport:
Currently if a UCR is not present in ILE then no notification is returned.
If a UCR is present but ILE is unresponsive for some reason, then an alert page must be displayed.
This is a Spike, therefore the page prototype was not yet available. A new ticket will be necessary to
update the alert page according to design guidelines.


app/views/consignment_not_found_page.scala.html
```diff
@@ -0,0 +1,36 @@
...
+@import views.Title
+@import views.Title.NO_SECTION
+
+@this(govukLayout: gds_main_template, pageTitle: pageTitle, sectionHeader: sectionHeader)
+
+@(ucr: String)(implicit request: Request[_], messages: Messages)
+
+@govukLayout(
+    title = Title("ileQueryResponse.ucrNotFound.title", NO_SECTION),
+    backButton = Some(BackButton(messages("site.back"), controllers.routes.ViewSubmissionsController.displayPage))) {
```
SteveSugden: Is this the right page to go back to?  Not the IleQueryController.displayQueryForm()?

app/controllers/ileQuery/IleQueryController.scala
```diff
@@ -115,7 +115,11 @@ class IleQueryController @Inject()(
 
         ducrResult.orElse(mucrResult).getOrElse(loadingPageResult)
 
-      case _: UcrNotFoundResponseExchangeData => Ok("UCR not found")
+      case response: UcrNotFoundResponseExchangeData =>
+        response.ucrBlock match {
+          case Some(UcrBlock(ucr, _)) => Redirect(controllers.ileQuery.routes.ConsignmentNotFoundController.displayPage(ucr))
```
SteveSugden: Did we need a whole new end-point and controller here?  Could we not just `Ok(consignmentNotFound(ucr))` (assuming the new not-found page was imported into this controller).

iservport: IMHO, this improves readability. The `IleQueryController` has already four other pages, namely  ile_query, loading_screen, ile_query_ducr_response, and ile_query_mucr_response.

app/views/consignment_not_found_page.scala.html
iservport: For sure. When I started the ticket, this page was not yet available. I will change it.

app/controllers/ileQuery/IleQueryController.scala
SteveSugden: IMO it makes the whole process even more complicated.  It's already bad enough with the redirects on checking the progress of the query.  The pages could be quite straight forward - 
1. start page (form)
2. in-progress page
3. two versions of "success"
4. failure (not found) page

iservport: I will remove the additional controller.

PR #110, opened 2020-02-04
CEDS-2149 - shut mucr option removed from choice page...

SteveSugden:
... if ILE query was for Ducr.
Fixed 'back' button on choice page to GDS

### DIFF
--- a/app/views/choice_page.scala.html
+++ b/app/views/choice_page.scala.html
@@ -86,7 +86,7 @@
                         content = Text(messages("movement.choice.submissions.label")),
                         checked = form("choice").value.contains(ViewSubmissions.value.toString)
                     )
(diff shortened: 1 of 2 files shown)


app/views/choice_page.scala.html
```diff
@@ -85,7 +86,7 @@
                         content = Text(messages("movement.choice.submissions.label")),
                         checked = form("choice").value.contains(ViewSubmissions.value.toString)
                     )
-                ),
+                ).filterNot(item => item.value.contains(ShutMUCR.value) && queryUcr.exists(_.ucrType.equals(UcrBlock.ducrType))),
```
bambuchaAdm: `==` instead of `equals` - it is not java :)

PR #113, opened 2020-02-11
CEDS-2100 - Make searching for a consignment the first step in the service

iservport:
The "Find a consignment" page must be the first step in the service of the internal movement so that NCH operators always search for a consignment before acting on it.

Important remarks:

1) The `hasIleQueryFeature` flag was removed,
2) The hidden routes previously in `testOnlyDoNotUseInAppConf.routes` are now in `app.routes`,
3) The `ReturnToStartException` redirects to the new "Find a consignment" page, and
4) The `Back Button` for several journeys has changed from the "Choice" page to the new "Find a consignment" page.

### DIFF
--- a/app/config/ErrorHandler.scala
+++ b/app/config/ErrorHandler.scala
@@ -36,7 +36,7 @@ class ErrorHandler @Inject()(val messagesApi: MessagesApi, errorTemplate: error)
   override def resolveError(request: RequestHeader, exception: Throwable): Result = exception match {
     case ReturnToStartException =>
       logger.warn(s"User Answers was in an invalid state, returning them to the Start Page from [${request.uri}]")
-      Results.Redirect(controllers.routes.ChoiceController.displayPage())
+      Results.Redirect(controllers.ileQuery.routes.IleQueryController.submitQueryForm())
     case _ =>
       logger.warn(s"Unexpected Exception was thrown accessing [${request.uri}]", exception)
       super.resolveError(request, exception)
--- a/app/config/ErrorHandler.scala
+++ b/app/config/ErrorHandler.scala
@@ -36,7 +36,7 @@ class ErrorHandler @Inject()(val messagesApi: MessagesApi, errorTemplate: error)
   override def resolveError(request: RequestHeader, exception: Throwable): Result = exception match {
     case ReturnToStartException =>
       logger.warn(s"User Answers was in an invalid state, returning them to the Start Page from [${request.uri}]")
-      Results.Redirect(controllers.ileQuery.routes.IleQueryController.submitQueryForm())
+      Results.Redirect(controllers.ileQuery.routes.IleQueryController.displayQueryForm())
     case _ =>
       logger.warn(s"Unexpected Exception was thrown accessing [${request.uri}]", exception)
       super.resolveError(request, exception)
--- a/app/config/ErrorHandler.scala
+++ b/app/config/ErrorHandler.scala
@@ -36,7 +36,7 @@ class ErrorHandler @Inject()(val messagesApi: MessagesApi, errorTemplate: error)
   override def resolveError(request: RequestHeader, exception: Throwable): Result = exception match {
     case ReturnToStartException =>
       logger.warn(s"User Answers was in an invalid state, returning them to the Start Page from [${request.uri}]")
-      Results.Redirect(controllers.ileQuery.routes.IleQueryController.displayQueryForm())
+      Results.Redirect(controllers.routes.ChoiceController.displayPage())
     case _ =>
       logger.warn(s"Unexpected Exception was thrown accessing [${request.uri}]", exception)
       super.resolveError(request, exception)
--- a/app/views/consignment_references.scala.html
+++ b/app/views/consignment_references.scala.html
@@ -32,7 +32,7 @@
     @helper.form(controllers.movements.routes.ConsignmentReferencesController.saveConsignmentReferences(), 'autoComplete -> "off") {
         @helper.CSRF.formField
 
-        @components.back_link(controllers.ileQuery.routes.IleQueryController.submitQueryForm())
+        @components.back_link(controllers.routes.ChoiceController.displayPage())
 
         @components.error_summary(form.errors)
 
--- a/app/views/consignment_references.scala.html
+++ b/app/views/consignment_references.scala.html
@@ -32,7 +32,7 @@
     @helper.form(controllers.movements.routes.ConsignmentReferencesController.saveConsignmentReferences(), 'autoComplete -> "off") {
         @helper.CSRF.formField
 
-        @components.back_link(controllers.routes.ChoiceController.displayPage())
+        @components.back_link(controllers.ileQuery.routes.IleQueryController.submitQueryForm())
 
         @components.error_summary(form.errors)
 
--- a/app/views/disassociate_ucr.scala.html
+++ b/app/views/disassociate_ucr.scala.html
@@ -28,7 +28,7 @@
         @helper.form(controllers.consolidations.routes.DisassociateUCRController.submit(), 'autoComplete -> "off") {
             @helper.CSRF.formField
 
-            @components.back_link(routes.ChoiceController.displayPage())
+            @components.back_link(controllers.ileQuery.routes.IleQueryController.submitQueryForm())
 
             @components.error_summary(form.errors)
 
--- a/app/views/disassociate_ucr.scala.html
+++ b/app/views/disassociate_ucr.scala.html
@@ -28,7 +28,7 @@
         @helper.form(controllers.consolidations.routes.DisassociateUCRController.submit(), 'autoComplete -> "off") {
             @helper.CSRF.formField
 
-            @components.back_link(controllers.ileQuery.routes.IleQueryController.submitQueryForm())
+            @components.back_link(routes.ChoiceController.displayPage())
 
             @components.error_summary(form.errors)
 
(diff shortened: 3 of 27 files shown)


app/config/ErrorHandler.scala
```diff
@@ -36,7 +36,7 @@ class ErrorHandler @Inject()(val messagesApi: MessagesApi, errorTemplate: error)
   override def resolveError(request: RequestHeader, exception: Throwable): Result = exception match {
     case ReturnToStartException =>
       logger.warn(s"User Answers was in an invalid state, returning them to the Start Page from [${request.uri}]")
-      Results.Redirect(controllers.routes.ChoiceController.displayPage())
+      Results.Redirect(controllers.ileQuery.routes.IleQueryController.submitQueryForm())
```
MaciejRewera: This is incorrect. AFAIR user should be redirected to `displayQueryForm` method in the same controller.

app/views/consignment_references.scala.html
```diff
@@ -32,7 +32,7 @@
     @helper.form(controllers.movements.routes.ConsignmentReferencesController.saveConsignmentReferences(), 'autoComplete -> "off") {
         @helper.CSRF.formField
 
-        @components.back_link(controllers.routes.ChoiceController.displayPage())
+        @components.back_link(controllers.ileQuery.routes.IleQueryController.submitQueryForm())
```
MaciejRewera: Why is this so? This way after choice page if I realise that I need to d osth else, I have to go through the whole ILE query again. I am quite sure this is not expected behaviour.

app/views/disassociate_ucr.scala.html
```diff
@@ -28,7 +28,7 @@
         @helper.form(controllers.consolidations.routes.DisassociateUCRController.submit(), 'autoComplete -> "off") {
             @helper.CSRF.formField
 
-            @components.back_link(routes.ChoiceController.displayPage())
+            @components.back_link(controllers.ileQuery.routes.IleQueryController.submitQueryForm())
```
MaciejRewera: Same here

app/views/ile_query.scala.html
```diff
@@ -15,26 +15,53 @@
...
+        govukInput: GovukInput,
+        errorSummary: errorSummary,
+        pageTitle: pageTitle,
+        formHelper: uk.gov.hmrc.play.views.html.helpers.FormWithCSRF
+)
 
 @(form: Form[String])(implicit request: Request[_], messages: Messages)
 
-@main_template(title = Title("ileQuery.title")) {
+@html = {
```
MaciejRewera: This variable does not explain what is it's purpose.

app/views/ile_query.scala.html
```diff
@@ -15,26 +15,53 @@
...
 
-    @components.page_title(Some(messages("ileQuery.title")))
+        @govukFieldset(Fieldset(
+            legend = Some(Legend(
+                content = Text(messages("ileQuery.title")),
+                classes = "govuk-fieldset__legend--xl",
+                isPageHeading = true
+            )),
+            html = html
+        ))
```
MaciejRewera: Please use components where you can. If I remember correctly in this situation you could use `exportsInputText`

app/views/ile_query.scala.html
```diff
@@ -15,26 +15,53 @@
...
-      field = form("ucr"),
-      label = ""
-    )
+        @govukButton(Button(content = Text(messages("site.continue"))))
 
-    @components.submit_button(messages("site.continue"))
-  }
+        <ul class="govuk-list">
+@*            TODO remove comments when CEDS-2156 is merged *@
+@*            <li>@link(message = messages("ileQuery.link.emptyMucr"), href = consolidations.routes.MucrOptionsController.displayPage())</li>*@
```
MaciejRewera: Nit: I don't think this is a part of this ticket. And if I am correct, even after removeing comments it would not work as you'd expect. It is redirecting to the first page of Associate journey.

app/views/view_submissions.scala.html
```diff
@@ -27,7 +27,7 @@
 
 @main_template(title = Title("submissions.title"), fullWidth = true) {
 
-  @components.back_link(routes.ChoiceController.displayPage())
+  @components.back_link(controllers.ileQuery.routes.IleQueryController.submitQueryForm())
```
MaciejRewera: 👍

test/unit/config/ErrorHandlerSpec.scala
```diff
@@ -69,7 +69,7 @@ class ErrorHandlerSpec extends UnitSpec with BeforeAndAfterEach with CSRFSupport
       val result = Future.successful(handler.resolveError(fakeRequest, ReturnToStartException))
 
       status(result) mustBe SEE_OTHER
-      redirectLocation(result) mustBe Some(controllers.routes.ChoiceController.displayPage().url)
+      redirectLocation(result) mustBe Some(controllers.ileQuery.routes.IleQueryController.submitQueryForm().url)
```
MaciejRewera: The test will pass but it is incorrect from business perspective.

app/views/consignment_references.scala.html
iservport: After a short discussion in the exports-design channel, I will revert this and create a back button pointing to the query page.

app/views/ile_query.scala.html
iservport: I agree, but in this case, the component you mentioned is slightly different as `exportsInputText ` does not render the title as in the prototype. Here is a comparison:

Screen Shot 2020-02-11 at 15 19 48

iservport:
For documentation purposes, these are the acceptance tests failing to be fixed:

 Failed tests:
[error]         spec.movements.internal.requests.notifications.InternalMovementNotificationsSpec
[error]         spec.movements.internal.arrivals.InternalArrivalsJourneySpec
[error]         spec.movements.internal.consolidations.disassociate.InternalDisassociateJourneySpec
[error]         spec.movements.internal.restrospectiveArrivals.RetroArrivalsJourneySpec
[error]         spec.movements.internal.consolidations.shutmucr.InternalShutMUCRJourneySpec
[error]         spec.movements.internal.requests.notifications.InternalConsolidationNotificationsSpec
[error]         spec.movements.internal.departures.InternalDeparturesJourneySpec
[error]         spec.movements.internal.consolidations.associate.InternalAssociateJourneySpec
[error]         spec.movements.internal.query.UcrSearchSpec

app/config/ErrorHandler.scala
iservport: Return to start must redirect to the original choice page.

app/views/ile_query.scala.html
iservport: I will refactor the component to add a class.

test/it/ChoiceSpec.scala
```diff
@@ -29,12 +29,12 @@ class ChoiceSpec extends IntegrationSpec {
...
     }
 
-    "return 200" in {
+    "return 303" in {
       givenAuthSuccess()
 
       val response = get(controllers.routes.ChoiceController.displayPage())
 
-      status(response) mustBe OK
+      status(response) mustBe SEE_OTHER
```
MaciejRewera: This is not a full solution, because now you have 2 cases: present the choice page if there is cache, or redirect if there is not.
Please add a test case to cover both scenarios.

test/unit/views/IleQueryViewSpec.scala
```diff
@@ -37,24 +39,47 @@ class IleQueryViewSpec extends ViewSpec {
...
-      view.getElementById("submit").text() mustBe messages("site.continue")
+      view.getElementsByClass("govuk-button").first().text() mustBe messages("site.continue")
+    }
+
+//    TODO better way to signal pending tests
+//    "contains link to create empty mucr" in {
+//
+//      govukListElement.getElementsByClass("govuk-link").get(0).text() mustBe messages("ileQuery.link.emptyMucr")
+//    }
+
```
MaciejRewera: Please remove this test case, as it is not going to be implemented any time shortly. There is no point to have commented test cases laying around.

test/unit/views/IleQueryViewSpec.scala
```diff
@@ -16,16 +16,18 @@
...
 
+import base.Injector
 import forms.IleQueryForm
+import org.jsoup.nodes.Element
 import play.api.mvc.{AnyContent, Request}
 import play.api.test.FakeRequest
 import views.html.ile_query
 
-class IleQueryViewSpec extends ViewSpec {
+class IleQueryViewSpec extends ViewSpec with Injector {
```
MaciejRewera: There is no test case for checking the input field.

MaciejRewera: And you can also add another one to test browser tab title.

test/unit/views/IleQueryViewSpec.scala
```diff
@@ -37,24 +40,46 @@ class IleQueryViewSpec extends ViewSpec {
...
+
+        govukErrorSummary.text() mustBe messages("error.summary.title")
       }
     }
 
+    val govukListElement = view.getElementsByClass("govuk-list").first()
+
+    "contains input field" in {
+
+//      view.getElementById("ucr") mustBe defined
```
MaciejRewera: Please uncomment

test/unit/views/IleQueryViewSpec.scala
```diff
@@ -37,24 +40,46 @@ class IleQueryViewSpec extends ViewSpec {
...
+
+        val errorView = page(IleQueryForm.form.withError("error", "error.required"))
+
+        val govukErrorSummary = errorView.getElementsByClass("govuk-error-summary__title").first()
+
+        govukErrorSummary.text() mustBe messages("error.summary.title")
       }
     }
 
+    val govukListElement = view.getElementsByClass("govuk-list").first()
```
MaciejRewera: Why is this outside test case?

PR #114, opened 2020-02-11
CEDS-2154 Sorting child ucrs based on the ROE status

PatrykRudnicki:


app/models/notifications/queries/RoeCode.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  case object NoControlRequired extends RoeCode {
+    val code = "6"
+    val priority = 4
+  }
+
+  case object UnknownRoe extends RoeCode {
+    val code = ""
+    val priority = 5
```
MaciejRewera: Here you can easily go with for e.g. 100. Won't need to rewrite it every time you add new code.

app/views/choice_page.scala.html
```diff
@@ -20,7 +20,7 @@
 @import forms.Choice._
 @import views.components.BackButton
 @import views.Title
-@import components.gds.{errorSummary, gds_main_template, pageTitle}
+@import components.gds.{childrenUcrs, errorSummary, gds_main_template, pageTitle}
```
SteveSugden: Why does the choice page need this import?

PatrykRudnicki: I checked before how the table looks like, this page is removed for now, I forgot to clean the import. Good spot!

app/models/notifications/EntryStatus.scala
```diff
@@ -16,10 +16,10 @@
 
 package models.notifications
 
-import models.notifications.queries.RoeCode
+import models.viewmodels.decoder.ROECode
 import play.api.libs.json.Json
 
-case class EntryStatus(ics: Option[String] = None, roe: Option[RoeCode] = None, soe: Option[String] = None)
+case class EntryStatus(ics: Option[String] = None, roe: Option[ROECode] = None, soe: Option[String] = None)
```
MaciejRewera: Nit: We've had a discussion some time ago not to capitalise codes. So better stay with `RoeCode`

PatrykRudnicki: ROECode is the class which already existed. RoeCode is the class that I created and removed in second comment. I'll rename this class and probably rest of the codes classes, but after rebasing to the Steve's branch.

If you look at the all changes you'll be able to see that I changed String to ROECode

app/models/viewmodels/decoder/ROECode.scala
```diff
@@ -16,24 +16,47 @@
...
+  case object UnknownRoe extends ROECode(code = "", messageKey = "", priority = 100)
+
 
-  case object DocumentaryControl extends ROECode(code = "1", messageKey = "decoder.roe.DocumentaryControl")
-  case object PhysicalExternalPartyControl extends ROECode(code = "2", messageKey = "decoder.roe.PhysicalExternalPartyControl")
-  case object NonBlockingDocumentaryControl extends ROECode(code = "3", messageKey = "decoder.roe.NonBlockingDocumentaryControl")
-  case object NoControlRequired extends ROECode(code = "6", messageKey = "decoder.roe.NoControlRequired")
-  case object RiskingNotPerformed extends ROECode(code = "0", messageKey = "decoder.roe.RiskingNotPerformed")
-  case object PrelodgePrefix extends ROECode(code = "H", messageKey = "decoder.roe.PrelodgePrefix")
+  implicit object ROECodeFormat extends Format[ROECode] {
```
SteveSugden: Why do we need the json formatter?

PatrykRudnicki: Because it's a part of the model that we're using for communication with the backend

PatrykRudnicki: This is why I said before it's slightly weird to reuse actual model, because viewmodels with formatter and everything can be not a good idea

But still probably there is more cons than maintaining two models

SteveSugden: I see.

PR #115, opened 2020-02-11
CEDS-2110 - change ile query response views to GDS and added additional

SteveSugden:
fields for parent and child consignment details


app/models/notifications/queries/AssociatedConsignmentsInfo.scala
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package models.notifications.queries
+
+case class AssociatedConsignmentsInfo(childDucrs: Seq[DucrInfo] = Seq.empty, childMucrs: Seq[MucrInfo] = Seq.empty) {
+
+  def size: Int = childDucrs.size + childMucrs.size
+
+  def mostSevereRoe: Option[String] =
```
SteveSugden: This is a temp. solution to extract a ROE - it's not the most severe and might not be needed at all if we show all of the children as a list

app/models/notifications/queries/MucrInfo.scala
```diff
@@ -25,7 +25,11 @@ case class MucrInfo(
   entryStatus: Option[EntryStatus] = None,
   isShut: Option[Boolean] = None,
   movements: Seq[MovementInfo] = Seq.empty
-)
+) {
+  def transport: Option[Transport] =
+    movements.find(movementInfo => movementInfo.transportDetails.isDefined).flatMap(movementWithTransport => movementWithTransport.transportDetails)
```
PatrykRudnicki: Probably you can replace `movementInfo` and `movementWithTransport` with `_`

app/models/notifications/queries/DucrInfo.scala
```diff
@@ -26,7 +26,12 @@ case class DucrInfo(
   entryStatus: Option[EntryStatus] = None,
   movements: Seq[MovementInfo] = Seq.empty,
   goodsItem: Seq[GoodsItemInfo] = Seq.empty
-)
+) {
+
+  def transport: Option[Transport] =
+    movements.find(movementInfo => movementInfo.transportDetails.isDefined).flatMap(movementWithTransport => movementWithTransport.transportDetails)
```
PatrykRudnicki: Probably you can replace `movementInfo` and `movementWithTransport` with `_`

app/views/TransportMapper.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import javax.inject.Singleton
+import models.notifications.queries.Transport
+import services.Countries
+import uk.gov.hmrc.govukfrontend.views.viewmodels.content.{Content, Empty, Text}
+
+@Singleton
+class TransportMapper {
+
+  private def countryName(code: String) = Countries.allCountries.find(_.countryCode == code).map(_.countryName).getOrElse(code)
```
PatrykRudnicki: This method can leave in the `Countries` object. There should be methods like this. I'm sure that in our different services you can find it, not sure if we have this in internal. If not it's better to put those methods there because it's general and can be reused

app/models/notifications/queries/MucrInfo.scala
SteveSugden: I find it easier to deal with mappings if I create a variable while I'm typing.  I usually then replace with "_" when possible.   I'll fix this.

app/views/TransportMapper.scala
SteveSugden: Good idea.  This mapper doesn't really do much.

PR #120, opened 2020-02-11
CEDS-2110 - Display consolidation information to NCH Officers

SteveSugden:
(This PR has grown to include CEDS-2154 - Display list of child consignments to NCH Officers)
- updated existing views to use GDS template and components
- added addition information regarding parent and child ucr
- incorporates changes to the ile query response model to facility sorting childs ucrs by "severity" of ROE code


app/models/notifications/EntryStatus.scala
```diff
@@ -16,9 +16,10 @@
 
 package models.notifications
 
+import models.viewmodels.decoder.ROECode
 import play.api.libs.json.Json
 
-case class EntryStatus(ics: Option[String] = None, roe: Option[String] = None, soe: Option[String] = None)
+case class EntryStatus(ics: Option[String] = None, roe: Option[ROECode] = None, soe: Option[String] = None)
```
MaciejRewera: Nit: We've had discussion some time ago to write `RoeCode`

app/models/notifications/queries/MucrInfo.scala
```diff
@@ -25,7 +25,11 @@ case class MucrInfo(
   entryStatus: Option[EntryStatus] = None,
   isShut: Option[Boolean] = None,
   movements: Seq[MovementInfo] = Seq.empty
-)
+) extends UcrInfo {
+  def transport: Option[Transport] =
+    movements.find(_.transportDetails.isDefined).flatMap(_.transportDetails)
```
MaciejRewera: This could be moved to `UcrInfo`

SteveSugden: Good idea.  The common trait didn't exist when I wrote this.   Keep the suggestions coming - I'll create a new PR

app/models/viewmodels/decoder/SOECode.scala
```diff
@@ -50,6 +50,8 @@ object SOECode {
 
   val MucrCodes: Set[SOECode] = Set(ConsolidationOpen, ConsolidationClosedWithoutP2P, ConsolidationHasP2P)
 
+  val CombinedCodes: Set[SOECode] = MucrCodes ++ DucrCodes
```
MaciejRewera: Nit: I'd call it `AllCodes` just because it raises less questions.

app/models/notifications/queries/IleQueryResponseExchangeData.scala
```diff
@@ -41,6 +42,8 @@ object IleQueryResponseExchangeData {
     childMucrs: Seq[MucrInfo] = Seq.empty
   ) extends IleQueryResponseExchangeData {
     override val typ = IleQueryResponseExchangeType.SuccessfulResponseExchange
+
+    val sortedChildrenUcrs: Seq[UcrInfo] = (childDucrs ++ childMucrs).sortBy(_.entryStatus.flatMap(_.roe).getOrElse(ROECode.UnknownRoe))
```
MaciejRewera: I don't think this is a good place to put it. Not under this name at least. The logic seems fine, but the name is very generic while the method is actually doing very specific type of sorting.
I'd rather put this sorting where it is used as it is a very specific case of sorting.

SteveSugden: I agree.   This should just be a method that returns the combine child ucr list.  The sorting should be done by the view.   Other users might have different sorting requirements.

app/views/IleCodeMapper.scala
```diff
@@ -15,31 +15,32 @@
...
 package views
+
+import javax.inject.{Inject, Singleton}
+import models.viewmodels.decoder.{CodeWithMessageKey, Decoder, ROECode}
 import play.api.i18n.Messages
+import uk.gov.hmrc.govukfrontend.views.viewmodels.content.{Content, Empty, HtmlContent, Text}
 
-object IleCodeMapper {
+@Singleton
+class IleCodeMapper @Inject()(decoder: Decoder) {
```
MaciejRewera: 👍 
Nit: Maybe a name that describes more specifically it's use case? Because this is returning a content specific to "Consignment Details" page

app/views/TransportMapper.scala
```diff
@@ -0,0 +1,35 @@
...
+@Singleton
+class TransportMapper {
+
+  def transportHtml(transport: Transport): Content =
+    (transport.transportId, transport.nationality) match {
+      case (Some(id), Some(nationality)) => Text(s"$id, ${countryName(nationality)}")
+      case (Some(id), None)              => Text(s"$id")
+      case (None, Some(nationality))     => Text(s"${countryName(nationality)}")
+      case _                             => Empty
+    }
```
MaciejRewera: Could it be just translating country (if exists) and combining with id?
I mean different implementation using for e.g. for comprehension? My first impresion was "Why so many cases?" ;)

SteveSugden: Both the id and nationality are optional and if they are both there I want a comma between them.  This was the first idea I had.  There may be others.

app/views/ile_query_ducr_response.scala.html
```diff
@@ -14,73 +14,87 @@
...
+                            content = Text(messages("ileQueryResponse.details.inputCustomsStatus"))
+                        ),
+                        value = Value(
+                            content = info.entryStatus.flatMap(_.ics).map(ileCodeMapper.inputCustomsStatus(_)).getOrElse(Empty)
+                        )
+                    )
+                ).filterNot(row => row.key.content == Text(messages("ileQueryResponse.details.transport")) && row.value.content == Empty),
+                classes = "govuk-!-margin-bottom-9"
+            ))
+        </div>
```
MaciejRewera: Can this part be also extracted into a component? Not to have multiple levels of abstraction here.

app/views/ile_query_mucr_response.scala.html
```diff
@@ -14,73 +14,91 @@
...
 
-@main_template(title = Title("ileQueryResponse.mucr.title")) {
-
-  @components.back_link(controllers.ileQuery.routes.IleQueryController.displayQueryForm())
-
-    @components.page_title(
-        Some(messages("ileQueryResponse.mucr.title")),
-        Some(info.ucr)
-    )
+        <div class="govuk-grid-column-two-thirds">
```
MaciejRewera: Why do you use grid definitions here? `govukLayout` already defines `tow-thirds` width for the page content.
Should you need a different width you can use `govukFlexibleLayout`

app/views/ile_query_mucr_response.scala.html
```diff
@@ -14,73 +14,91 @@
...
 
-@this(main_template: views.html.templates.main_template)
+@this(
+        ileCodeMapper: IleCodeMapper,
+        transportMapper: TransportMapper,
+        govukLayout: gds_main_template,
+        parentSection: response_parent,
+        previousMovementsSection: response_previous_movements,
+        associatedConsignmentsSection: response_associated_consignments,
+        responseButtons: response_buttons,
```
MaciejRewera: Nit: IMHO it would be better to use 'result' instead of 'response'. To me response suggests the payload itself.
Might be also worth asking if it would be better to align views names with pages titles?

conf/messages.en
```diff
@@ -321,60 +321,28 @@ ileQueryResponse.ducr.title = Consignment information
...
-
-ileQuery.mapping.soe.mucr.default = Unknown Status of Entry ({0})
-ileQuery.mapping.soe.mucr.0 = The MUCR is still open, or contains no declarations.
-ileQuery.mapping.soe.mucr.C = Consolidation is closed but does not have P2P
-ileQuery.mapping.soe.mucr.3 = Consolidation is ready to go
-
-
-ileQuery.mapping.ics.default = Unknown Input Customs Status ({0})
-ileQuery.mapping.ics.3 = Declaration invalidated by trader
-ileQuery.mapping.ics.6 = Declaration invalidated by customs
```
MaciejRewera: 👍 Thank you!

app/views/ile_query_mucr_response.scala.html
SteveSugden: The tables need to be full width.  I copied the example of "movements" page in external service.

test/unit/controllers/ileQuery/IleQueryControllerSpec.scala
```diff
@@ -109,7 +111,7 @@ class IleQueryControllerSpec extends ControllerLayerSpec with MockIleQueryCache
         val result = controller.submitQuery("mucr")(request)
 
         status(result) mustBe OK
-        verify(ileQueryMucrResponsePage).apply(meq(mucrInfo))(any(), any())
+        verify(ileQueryMucrResponsePage).apply(meq(mucrInfo), meq(parentMucrInfo), meq(Seq.empty))(any(), any())
```
MaciejRewera: Is it worth checking child UCRs being passed ?

test/unit/models/notifications/queries/IleQueryResponseExchangeDataSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+
+      val response = SuccessfulResponseExchangeData(childDucrs = childDucrs, childMucrs = childMucrs)
+
+      val expectedStatusOrder =
+        Seq(PhysicalExternalPartyControl, DocumentaryControl, NonBlockingDocumentaryControl, NoControlRequired)
+
+      response.sortedChildrenUcrs.map(_.entryStatus.flatMap(_.roe).get) mustBe expectedStatusOrder
+    }
+  }
+}
```
MaciejRewera: As per my previous comment on the placement of this method, this could be removed.

test/unit/models/viewmodels/notificationspage/converters/EMRResponseConverterSpec.scala
```diff
@@ -142,14 +142,14 @@ class EMRResponseConverterSpec extends UnitSpec with BeforeAndAfterEach {
         contentBuilder.convert(input)
 
         verify(decoder).crc(meq(UnknownCrcCode))
-        verify(decoder).roe(meq(UnknownRoeCode))
+        verify(decoder).roe(meq(UnknownRoeCode.code))
```
MaciejRewera: Why only one of those needs `.code` ?

test/unit/models/viewmodels/notificationspage/converters/EMRResponseConverterSpec.scala
```diff
@@ -191,12 +191,11 @@ object EMRResponseConverterSpec {
...
     messageCode = EMR.code,
     timestampReceived = testTimestamp,
-    entries =
-      Seq(Entry(ucrBlock = Some(UcrBlock(ucr = correctUcr, ucrType = "M")), entryStatus = Some(EntryStatus(roe = Some(roeKeyFromDecoder.code)))))
+    entries = Seq(Entry(ucrBlock = Some(UcrBlock(ucr = correctUcr, ucrType = "M")), entryStatus = Some(EntryStatus(roe = Some(roeKeyFromDecoder)))))
   )
 
   val UnknownCrcCode = "1234"
-  val UnknownRoeCode = "456"
+  val UnknownRoeCode = ROECode.UnknownRoe
```
MaciejRewera: I'd add `.code` here. To be concise.

test/unit/models/viewmodels/notificationspage/converters/ERSResponseConverterSpec.scala
```diff
@@ -195,7 +195,7 @@ object ERSResponseConverterSpec {
   )
 
   val UnknownIcsCode = "123"
-  val UnknownRoeCode = "456"
+  val UnknownRoeCode = ROECode.UnknownRoe
```
MaciejRewera: Same here.

test/unit/views/IleQueryDucrResponseViewSpec.scala
```diff
@@ -18,33 +18,40 @@ package views
...
-  val status = EntryStatus(Some("ICS"), Some("ROE"), Some("SOE"))
+  val status = EntryStatus(Some("ICS"), None, Some("SOE"))
   val ducrInfo =
     DucrInfo(ucr = "8GB123458302100-101SHIP1", declarationId = "121332435432", movements = Seq.empty, entryStatus = Some(status))
 
-  private def view(info: DucrInfo = ducrInfo) = page(info)
+  private def view(info: DucrInfo = ducrInfo, parent: Option[MucrInfo] = None) = page(info, parent)
+
+  private def summaryElement(html: Html, index: Int) = html.getElementById("summary").select(s"div:eq($index)>dd").get(0)
+  private def parentConsignmentElement(html: Html, index: Int) = html.getElementById("parentConsignment").select(s"div:eq($index)>dd").get(0)
```
MaciejRewera: Can it be selected by a class name?

test/unit/views/IleQueryDucrResponseViewSpec.scala
```diff
@@ -18,33 +18,40 @@ package views
...
+    MovementInfo(messageCode = "RET", goodsLocation = "GBAUDFGFSHFKD", movementDateTime = Some(ZonedDateTime.parse("2019-11-04T16:27:18Z").toInstant))
   val depart = MovementInfo(
     messageCode = "EDL",
     goodsLocation = "GBAUFDSASFDFDF",
-    movementDateTime = Some(ZonedDateTime.parse("2019-10-30T12:34:18Z").toInstant)
+    movementDateTime = Some(ZonedDateTime.parse("2019-10-30T09:17:18Z").toInstant)
   )
 
-  val status = EntryStatus(Some("ICS"), Some("ROE"), Some("SOE"))
+  val status = EntryStatus(Some("ICS"), None, Some("SOE"))
```
MaciejRewera: Why was it changed to `None` ?

test/unit/views/IleQueryDucrResponseViewSpec.scala
```diff
@@ -18,33 +18,40 @@ package views
...
+import models.notifications.queries.{DucrInfo, MovementInfo, MucrInfo}
+import models.viewmodels.decoder.ROECode.UnknownRoe
+import models.viewmodels.decoder.{ICSCode, ROECode, SOECode}
 import play.api.mvc.{AnyContent, Request}
 import play.api.test.FakeRequest
+import play.twirl.api.Html
 import views.html.ile_query_ducr_response
 
-class IleQueryDucrResponseViewSpec extends ViewSpec {
+class IleQueryDucrResponseViewSpec extends ViewSpec with Injector {
```
MaciejRewera: There is a lot of tests in here. Might be worth creating tests for the components (page parts) and then only check if this view calls them with correct parameters.

test/unit/views/IleQueryMucrResponseViewSpec.scala
```diff
@@ -18,33 +18,41 @@ package views
...
+import models.notifications.queries.{MovementInfo, MucrInfo, UcrInfo}
+import models.viewmodels.decoder.ROECode.UnknownRoe
+import models.viewmodels.decoder.{ROECode, SOECode}
 import play.api.mvc.{AnyContent, Request}
 import play.api.test.FakeRequest
+import play.twirl.api.Html
 import views.html.ile_query_mucr_response
 
-class IleQueryMucrResponseViewSpec extends ViewSpec {
+class IleQueryMucrResponseViewSpec extends ViewSpec with Injector {
```
MaciejRewera: Here I have the same questions/issues as for the IleQueryDucrResponseViewSpec

app/views/ile_query_mucr_response.scala.html
MaciejRewera: Yes, you're right. I forgot how it was working.

app/views/TransportMapper.scala
PatrykRudnicki:
```
def transportHtml(transport: Transport): Content =
    Text((transport.transportId.toSeq ++ transport.nationality.map(countryName(_)).toSeq).mkString(", "))
```

PR #126, opened 2020-02-13
BAU - move creation of IleQueries ttl index to mongobee changelog

SteveSugden:
This change is not strictly necessary but it keeps the creation of indexes consistent (i.e. in the mongobee changelog scripts).

In addition, it solves the problem that the ttl timeout value was not changeable via the config (if the db already had the index).


SteveSugden:
Good point.  I'll check

PR #137, opened 2020-02-17
Poc - trying Mongock as alternative to Mongobee

SteveSugden:


project/AppDependencies.scala
```diff
@@ -16,7 +16,7 @@ object AppDependencies {
     "uk.gov.hmrc"          %% "simple-reactivemongo"          % "7.22.0-play-26",
     "ai.x"                 %% "play-json-extensions"          % "0.40.2",
     "com.github.tototoshi" %% "scala-csv"                     % "1.3.6",
-    "com.github.mongobee"  %  "mongobee"                      % "0.13"
+    "com.github.cloudyrock.mongock"  %  "mongock-core"                      % "2.0.2"
```
SteveSugden: My IDE doesn't format this file correctly.  :-(

app/mongock/changesets/CacheChangelog.java
```diff
@@ -0,0 +1,39 @@
...
+import java.time.Instant;
+import java.util.Date;
+
+@ChangeLog
+public class CacheChangelog {
+    private String collection = "cache";
+
+    @ChangeSet(order = "001", id = "Internal Movements DB Baseline", author = "Steve Sugden")
+    public void dbBaseline(MongoDatabase db) {
+        Document query = new Document();
```
SteveSugden: If this works, it will add a new field to existing cache documents (that doesn't exist in the model).
We can then safely delete lines 34-36.

app/config/AppConfig.scala
```diff
@@ -28,6 +29,10 @@ class AppConfig @Inject()(
   @Named("appName") serviceIdentifier: String
 ) {
 
+  runModeConfiguration
+    .getOptional[String]("mongodb.uri")
+    .map(uri => MongockConfig(uri))
+
```
PatrykRudnicki: Should we have some logic to handle missing configuration? I mean some `.getOrElse(throw Exception`?

SteveSugden: I did this deliberately as I didn't want to add "mongodb.uri" config to the tests.    If this config value is missing then we have bigger problems than not running the upgrade scripts.

PR #142, opened 2020-02-18
BAU - disable mongock to unblock pipeline

SteveSugden:


app/config/AppConfig.scala
```diff
@@ -29,9 +29,10 @@ class AppConfig @Inject()(
   @Named("appName") serviceIdentifier: String
 ) {
 
-  runModeConfiguration
-    .getOptional[String]("mongodb.uri")
-    .map(uri => MongockConfig(uri))
+//  // TODO - fix or remove
```
PauloPortugal: Should we try to update the latest `java-mongo-driver`?

SteveSugden: We can try all sorts of things.  But after we get the pipeline unblocked.   There's stuff already deployed to QA that is wrong.

PR #145, opened 2020-02-19
CEDS-2177 - Improvements to Queried UCR page

SteveSugden:
- Omit table rows with no values (codes)
- Hide movements table when empty
- Move submit button & link up under Queried UCR section

- Fixed bug whereby unknown route of entry was displayed as "Unknown"
rather than "<code> - Unknown"

- Took the opportunity to break up the existing view specs in favour of
smaller tests for each component.

### DIFF
--- a/app/models/viewmodels/decoder/ROECode.scala
+++ b/app/models/viewmodels/decoder/ROECode.scala
@@ -43,7 +43,7 @@ object ROECode {
       NoControlRequired,
       RiskingNotPerformed,
       PrelodgePrefix,
-      UnknownRoe
+      UnknownRoe()
     )
 
   case object DocumentaryControl extends ROECode(code = "1", messageKey = "decoder.roe.DocumentaryControl", priority = 2)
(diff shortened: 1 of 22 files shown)


app/models/viewmodels/decoder/ROECode.scala
```diff
@@ -43,7 +43,7 @@ object ROECode {
       NoControlRequired,
       RiskingNotPerformed,
       PrelodgePrefix,
-      UnknownRoe
+      UnknownRoe()
```
PatrykRudnicki: I think you should add here `NoneRoe` too

SteveSugden: Why?   Its not a code that needs to be decoded

PatrykRudnicki: Yeah, you're right! My fault

SteveSugden: If anything, perhaps we should remove 'UnknownRoe' and rename to 'validCodes' or 'kniwnCodes'

PR #151, opened 2020-02-24
CEDS-2170 ile query controller fix

MaciejRewera:


app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -108,8 +106,8 @@ class CustomsDeclareExportsMovementsConnector @Inject()(appConfig: AppConfig, ht
     httpClient
       .GET[HttpResponse](s"${appConfig.customsDeclareExportsMovementsUrl}$IleQuery/$conversationId", providerIdQueryParam(providerId))
       .recoverWith {
-        case exception: Upstream4xxResponse if exception.upstreamResponseCode == FailedDependencyStatus =>
-          Future.successful(HttpResponse(responseStatus = FailedDependencyStatus))
+        case exception: Upstream4xxResponse if exception.upstreamResponseCode == Status.FAILED_DEPENDENCY =>
```
PauloPortugal: :+1:

app/models/notifications/queries/IleQueryResponseExchangeData.scala
```diff
@@ -44,6 +44,16 @@ object IleQueryResponseExchangeData {
     override val typ = IleQueryResponseExchangeType.SuccessfulResponseExchange
 
     val sortedChildrenUcrs: Seq[UcrInfo] = (childDucrs ++ childMucrs).sortBy(_.entryStatus.flatMap(_.roe).getOrElse(ROECode.NoneRoe))
+
+    lazy val queriedUcr: UcrInfo = {
```
SteveSugden: You can do something like this (based on the controller code you are removing)
```
lazy val queriedUcr: UcrInfo = queriedDucr
      .orElse(queriedMucr)
      .getOrElse(throw new IllegalStateException("SuccessfulResponseExchangeData must have either queriedDucr or queriedMucr"))
```

SteveSugden: Can the `sortedChildrenUcrs` val be lazy as well?

app/models/notifications/queries/UcrInfo.scala
```diff
@@ -24,5 +26,10 @@ abstract class UcrInfo {
   val entryStatus: Option[EntryStatus]
   val movements: Seq[MovementInfo]
 
-  def transport: Option[Transport] = movements.find(_.transportDetails.isDefined).flatMap(_.transportDetails)
+  def transport: Option[Transport] =
```
SteveSugden: can this be a lazy val also - no need to run this expensive computation more than once.

PR #159, opened 2020-03-02
CEDS-2178 confirmation pages enhancements

MaciejRewera:
In this PR I've changed the layout of the Confirmation Pages. In the end I decided to create a single confirmation_page that is re-used for every journey. The only difference between journeys is the text, so it naturally makes sense to use a single twirl template.

The other part of the PR is to make it impossible to access Confirmation Page directly. This is done by adding Flash header in Redirect response in Summary Controllers.

The last part is refactoring. Except removing/moving around some classes, I've better organised view files and corresponding specs to mirror each other location.


app/controllers/movements/MovementConfirmationController.scala
```diff
@@ -21,21 +21,31 @@ import controllers.storage.FlashKeys
...
-    Ok(page(`type`))
+  def displayPage: Action[AnyContent] = authenticate { implicit request =>
+    val journeyType = extractJourneyType
+    journeyType match {
+      case ARRIVE | RETROSPECTIVE_ARRIVE | DEPART => Ok(confirmationPage(journeyType))
+      case _                                      => throw ReturnToStartException
+    }
   }
 
+  private def extractJourneyType(implicit request: Request[_]): JourneyType =
```
SteveSugden: Similar to external movements - maybe move this method to trait?

PR #160, opened 2020-03-03
CEDS-2129 Introduce GDS design system to Depart journey on Internal Movements

janeturner:
Use GDS on Departure Details Page
Use GDS on Goods Departed Page
Use GDS on Transport Page
Use GDS on Departure Summary page


app/controllers/movements/TransportController.scala
```diff
@@ -44,19 +45,22 @@ class TransportController @Inject()(
 
   def displayPage(): Action[AnyContent] = (authenticate andThen getJourney(JourneyType.DEPART)) { implicit request =>
     val answers = request.answersAs[DepartureAnswers]
+    val consignmentReference = answers.consignmentReferences.map(_.referenceValue).getOrElse(throw ReturnToStartException)
```
bambuchaAdm: `.map(_.referenceValue).getOrElse(throw ReturnToStartException)` is quite repeating pattern over code. Maybe would be worth to extract it as method to answers/references class ?

app/views/transport.scala.html
```diff
@@ -14,58 +14,125 @@
...
 
-        @components.back_link(routes.GoodsDepartedController.displayPage())
+    @formHelper(action = controllers.movements.routes.TransportController.saveTransport(), 'autoComplete -> "off") {
+        @errorSummary(form.errors)
 
-        @components.error_summary(form.errors)
+        @sectionHeader(messages("transport.heading", consignmentReference))
 
-        @components.page_title(Some(messages("transport.title")))
+        @pageTitle(messages("transport.title"))
```
bambuchaAdm: Does page title has to be in form ?

PR #163, opened 2020-03-09
CEDS-2199 - Prevent changes to UCR when Managing Consignment

iservport:
**Shut Mucr page change button removed.**

The Internal and External Movements Services are being changed to force users to query before performing an action in order to reduce the risk that a movement is inadvertently performed against the wrong UCR. Therefore, it doesn't make sense to give users the option to change the UCR when managing the queried consignment.

After searching for a valid UCR and proceeding to the next page via  'Manage this Consignment' button, following any action to perform against this UCR (e.g. Arrive, Depart, Shut, Disassociate, Retrospective Arrival), it must be not possible to change the UCR.

### DIFF
--- a/app/views/shutmucr/shut_mucr_summary.scala.html
+++ b/app/views/shutmucr/shut_mucr_summary.scala.html
@@ -44,16 +44,7 @@
                         ),
                         value = Value(
                             content = Text(mucr)
-                        ),
-                        actions = Some(Actions(
-                            items = Seq(
-                                ActionItem(
-                                    href = s"${controllers.ileQuery.routes.FindConsignmentController.displayQueryForm()}",
-                                    content = Text(messages("site.change")),
-                                    visuallyHiddenText = Some(messages("site.change.hint.disassociate.ucr", mucr.toUpperCase()))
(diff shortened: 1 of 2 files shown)


app/views/shutmucr/shut_mucr_summary.scala.html
```diff
@@ -44,16 +44,7 @@
                         ),
                         value = Value(
                             content = Text(mucr)
-                        ),
```
PauloPortugal: Shouldn't this change be based on a feature flag?
By removing here, we will lose the ability to change it for all journeys (with feature flag enabled and disabled)

iservport: It was agreed with Ian to remove the change button from all journeys on the internal frontend. Arrive, Depart, etc. have already been updated, this one was forgotten. Additionally, if the user needs to change the consignment, the back button lets her/him find a new consignment from the query (initial) page.

PauloPortugal: Cool! :) Thank you for the info

PR #164, opened 2020-03-09
CEDS-2214 - Internal Movements: Add MRN to Movements Query results page

iservport:
A queried DUCR must show the MRN on the results page. In contrast, there must be no MRN displayed on the results page when a MUCR is queried.

### DIFF
--- a/app/views/components/ilequery/response_ducr_summary.scala.html
+++ b/app/views/components/ilequery/response_ducr_summary.scala.html
@@ -62,6 +62,14 @@
                 value = Value(
                     content = info.entryStatus.flatMap(_.ics).map(converter.inputCustomsStatus(_)).getOrElse(Empty)
                 )
+            ),
+            SummaryListRow(
+                key = Key(
+                    content = Text(messages("ileQueryResponse.details.MRN"))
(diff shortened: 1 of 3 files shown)


app/views/components/ilequery/response_ducr_summary.scala.html
```diff
@@ -62,6 +62,14 @@ <h2 class="govuk-heading-m">@messages("ileQueryResponse.details")</h2>
                 value = Value(
                     content = info.entryStatus.flatMap(_.ics).map(converter.inputCustomsStatus(_)).getOrElse(Empty)
                 )
+            ),
+            SummaryListRow(
+                key = Key(
+                    content = Text(messages("ileQueryResponse.details.MRN"))
+                ),
+                value = Value(
+                    content = HtmlContent(info.declarationId)
```
bambuchaAdm: Are you sure that is not our internal id ? MRN ?

iservport: Yes, it is stated in the ticket, and to double-check, I asked Ian as well.

PR #166, opened 2020-03-09
CEDS-2130 retro arrival govuk design system

MaciejRewera:
The only page that required updating was the Summary page.


app/views/summary/arrival_summary_page.scala.html
```diff
@@ -80,33 +80,15 @@
...
-                        actions = Some(Actions(
-                            items = Seq(
-                                ActionItem(
-                                    href = s"${controllers.movements.routes.MovementDetailsController.displayPage}",
-                                    content = Text(messages("site.change")),
-                                    visuallyHiddenText = Some(messages("summary.arrival.date.change"))
-                                )
-                            )
-                        ))
+                        )
```
SteveSugden: Just curious; why no change links?

MaciejRewera: Because they are not present on other summary pages. Wasn't there a ticket to remove those?
Arrival Summary page haven't had view tests for some time, so it might have been omitted.

test/unit/views/movement/ArrivalDetailsViewSpec.scala
```diff
@@ -27,40 +35,154 @@ class ArrivalDetailsViewSpec extends ViewSpec with Injector {
...
 
-    "render heading input with hint for time" in {
-      page(movementDetails.arrivalForm(), None).getElementById("timeOfArrival-hint") must containMessage("arrivalDetails.time.hint")
+    "provided with form containing data" should {
+      val date = LocalDate.now().minusDays(1)
+      val time = LocalTime.of(1, 2)
+      val viewWithData = createView(movementDetails.arrivalForm().fill(ArrivalDetails(Date(date), Time(time))))
+
+      "have value in day field" in {
+        viewWithData.getElementById("dateOfArrival_day").`val`() mustBe convertIntoTwoDigitFormat(date.getDayOfMonth)
```
SteveSugden: Why didn't you just hard-code the expected 2/4 digit values?  I can't see any tests for the two `convert` methods.

app/views/summary/arrival_summary_page.scala.html
SteveSugden: Are you sure?  I've just tried Arrival in internal and external service.  The date/time and location have change links

MaciejRewera: Spoke to the BAs about t and you're right. So there was an error on Arrival Summary page, that 'change' links were not present while they should. fixing this just now.

test/unit/views/movement/ArrivalDetailsViewSpec.scala
MaciejRewera: These methods are very simple and located at the beginning of the file. The question is not to me actually, because I just copied these into a proper place.
And you cannot hard-code the values, because the validation is date-dependent, so the input data is generated new every execution.

SteveSugden: You could create the test dates with known values rather than relying on "now".    I know it won't make a difference in this case, but tests using current date/time can be a recipe for failing at some point in the future (usually when transitioning into or out of daylight savings).

Anyway, no change required.  Happy to approve once the 'change' links are re-inrtoduced.

PR #167, opened 2020-03-11
CEDS-2130 retro arrival location fix

MaciejRewera:
This PR fixes Section Header on Location page for Retrospective Arrival.
Additionally, a section header has been added to Goods Location page.

### DIFF
--- a/test/unit/controllers/movements/GoodsDepartedControllerSpec.scala
+++ b/test/unit/controllers/movements/GoodsDepartedControllerSpec.scala
@@ -17,11 +17,11 @@
 package controllers.movements
 
 import controllers.ControllerLayerSpec
-import forms.GoodsDeparted
 import forms.GoodsDeparted.DepartureLocation.BackIntoTheUk
+import forms.{ConsignmentReferenceType, ConsignmentReferences, GoodsDeparted}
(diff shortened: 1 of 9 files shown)


test/unit/controllers/movements/GoodsDepartedControllerSpec.scala
```diff
@@ -36,13 +36,15 @@ class GoodsDepartedControllerSpec extends ControllerLayerSpec with MockCache {
...
+  private val consignmentReferences = ConsignmentReferences(ConsignmentReferenceType.D, "referenceValue")
+
+  private def controller(answers: Answers = DepartureAnswers(consignmentReferences = Some(consignmentReferences))) =
     new GoodsDepartedController(SuccessfulAuth(), ValidJourney(answers), cacheRepository, stubMessagesControllerComponents(), page)(global)
 
   override def beforeEach(): Unit = {
     super.beforeEach()
 
-    when(page.apply(any())(any(), any())).thenReturn(HtmlFormat.empty)
+    when(page.apply(any(), anyString())(any(), any())).thenReturn(HtmlFormat.empty)
```
PatrykRudnicki: Not sure if using `anyString` make any value. If the field is as String type, any is enough because in general you're not able to put anything else.

Always to avoid more imports we can go with `any[String]`.

But this is correct

PR #170, opened 2020-04-07
BAU Unify AssociateKind and DisassociateKind into UcrType

MaciejRewera:
Both _Kind classes were essentially the same so I decided
to merge these into a single type.


MaciejRewera:
This PR requires some fixes first. It was opened too early.

PR #172, opened 2020-04-14
CEDS-2200 No UCR change on Associate

MaciejRewera:


app/controllers/ChoiceController.scala
```diff
@@ -50,31 +51,44 @@ class ChoiceController @Inject()(
...
+      val redirectionCall = cache.queryUcr
+        .map(_.ucrType match {
+          case Ducr.codeValue => consolidations.routes.MucrOptionsController.displayPage()
+          case Mucr.codeValue => consolidations.routes.ManageMucrController.displayPage()
+        })
+        .getOrElse(throw ReturnToStartException)
+
+      saveAndRedirect(AssociateUcrAnswers.fromQueryUcr, redirectionCall)
+    }
+
```
PatrykRudnicki, line 87: Unnecessary brackets, but it can be more readable for someone.

app/models/UcrType.scala
```diff
@@ -0,0 +1,46 @@
...
+      data.get(key).map { typ =>
+        lookup.andThen(Right.apply).applyOrElse(typ, (_: String) => Left(Seq(FormError(key, "error.unknown"))))
+      }
+    }.getOrElse(Left(Seq(FormError(key, "error.required"))))
+
+    override def unbind(key: String, value: UcrType): Map[String, String] = Map(key -> value.formValue)
+  }
+
+  implicit val format =
+    Format[UcrType](Reads.StringReads.collect(JsonValidationError("error.unknown"))(lookup), Writes(kind => JsString(kind.formValue)))
```
PatrykRudnicki, line 45: I think it's better to use different name instead of `kind`. Something more related with `UcrType`

conf/app.routes
```diff
@@ -12,27 +12,31 @@ GET         /choice                                controllers.ChoiceController.
 GET         /choice/:journey                       controllers.ChoiceController.startSpecificJourney(journey: forms.Choice)
 POST        /choice                                controllers.ChoiceController.submit()
 
+#Manage MUCR
+GET         /manage-mucr                          controllers.consolidations.ManageMucrController.displayPage()
+POST        /manage-mucr                          controllers.consolidations.ManageMucrController.submit()
```
PatrykRudnicki, line 17: Why we have a separate endpoints for manage MUCR? What is the reason of this page?

Just asking, because not sure if it's related with this blocking the ucr field. Probably I don't understand what is happening and what is the reason of this Controller and endpoints

app/models/UcrType.scala
PatrykRudnicki: And why we're just saving the value in the writes?

app/controllers/ChoiceController.scala
MaciejRewera: That was the purpose.

app/models/UcrType.scala
MaciejRewera: Because if I were to store the whole object I'd need to update Backend, other Frontend and perform DB migrations.
This is the tech-debt we have because we have not looked at it wholistically and we have several ways to handle UCR and UCR type.

conf/app.routes
MaciejRewera: For details, please have a look at the ticket or the prototype.
In short: this page needs to be present if user queried for MUCR in order to determine queried MUCR as parent or child MUCR. Depending on the answer the user is redirected to different page.

MaciejRewera:
Because of doing how you just proposed we got ourselves in a position that depending on where the UCR is used we encode UCR type differently (for e.g. 'ducr' or 'D').
With `UcrType` I've merged `AssociateKind` and `DisassociateKind` but I could not merge-in `ConsignmentReference` as they are using just D or M and therefore JSON formats for `UcrType` would need to handle 2 scenarios.

It is funny, because the problem was there the whole time, but only when I optimised the code it became visible.
The reason there are so many Ucr-like classes is because we were implementing them in separation, not knowing what someone else is doing (excapt `UcrBlock` which is a copy from ILE model).
I agree that this implementation is not perfect but let's treat it as a cornerstone for improving how we approach common parts of our domain.

PatrykRudnicki:
@MaciejRewera Your first commit is not signed, and the merging is blocked.

MaciejRewera:
@SteveSugden Refactoring for `UcrType` can be done independently from feature switch. It is a good idea to align these two.

PR #175, opened 2020-05-18
CEDS-2391 Fix for timeoutDialog during the journey

PatrykRudnicki:


app/views/components/gds/gds_main_template.scala.html
```diff
@@ -40,7 +40,7 @@
   useCustomContentWidth: Boolean = false
 )(contentBlock: Html)(implicit request: Request[_], messages: Messages)
 
-@useTimeoutDialog = @{ request.isInstanceOf[AuthenticatedRequest[_]]}
+@useTimeoutDialog = @{ request.isInstanceOf[AuthenticatedRequest[_]] || request.isInstanceOf[JourneyRequest[_]] }
```
SteveSugden, line 43: I can see why you are making this change.  I'm not sure how this fix is not required in customs-declare-frontend

PatrykRudnicki: Journey Request in customs-declare-exports-frontend extends AuthenticatedRequest

https://github.com/hmrc/customs-declare-exports-frontend/blob/master/app/models/requests/JourneyRequest.scala

PatrykRudnicki: In internal service journeyRequest just contains the authenticatedRequest inside

PR #180, opened 2020-06-18
CEDS-2482 - fix link on service name in header

SteveSugden:


SteveSugden:
No feature toggle for internal service.   I did check because I couldn't remember which service had the toggle

PR #187, opened 2020-07-14
CEDS-2607 Libraries upgrade

PatrykRudnicki:


app/connectors/CustomsDeclareExportsMovementsConnector.scala
```diff
@@ -105,10 +106,6 @@ class CustomsDeclareExportsMovementsConnector @Inject()(appConfig: AppConfig, ht
   def fetchQueryNotifications(conversationId: String, providerId: String)(implicit hc: HeaderCarrier): Future[HttpResponse] =
     httpClient
       .GET[HttpResponse](s"${appConfig.customsDeclareExportsMovementsUrl}$IleQuery/$conversationId", providerIdQueryParam(providerId))
-      .recoverWith {
```
PatrykRudnicki, line 108: As in `customs-movements-frontend` this part is not necessary. It takes response with failed dependency and create response with failed dependency.

Whole logic to handle it is in controller.

PR #190, opened 2020-07-17
CEDS-2531 - hint full-stops

SteveSugden:


SteveSugden:
Screen readers combine the hint with a field-level error message and without a full-stop they run into each other, apparently.

PR #191, opened 2020-07-17
CEDS-2634 Display time in 12 hours format

PatrykRudnicki:


app/views/ViewDates.scala
```diff
@@ -18,10 +18,22 @@ package views
 
 import java.time.ZoneId
 import java.time.format.DateTimeFormatter
+import java.time.temporal.TemporalAccessor
 
-object ViewDates {
+import javax.inject.{Inject, Singleton}
+
+@Singleton
+class ViewDates @Inject()() {
```
SteveSugden, line 26: What are you injecting?

PatrykRudnicki: Nothing here, this give you possibility to inject ViewDates in other places

SteveSugden: So it makes the class injectable without having to register it with a module?  Nice.
Should you not remove the DateTimeFormatterProvider from DateTimeModule?  In case it is being used anywhere else?

PatrykRudnicki: yeah, I'll remove this

MaciejRewera: @PatrykRudnicki You don't need `@Inject` annotation to be able to inject the class into another one (but it has to be `class`, not `object`). You can inject any class you like, for e.g. `Decoder` is injected in many Converter classes, but it has no `@Inject` annotation.
`DateTimeFormatterProvider` was enough some time ago, but since we have a dedicated class for this purpose, it makes sense to remove it.

PatrykRudnicki: Ah, I thought that Inject is necessary.

Thanks, I'll change it on Monday 🙂

SteveSugden: I've changed it on my movements branch and it still works!

PR #195, opened 2020-07-31
CEDS 2640 Add Ducr Part journey

PatrykRudnicki:
- CEDS-2640 Add Ducr Part journey
- CEDS-2640 Add feature switch to service and DUCR Part feature

### DIFF
--- a/app/models/cache/Answers.scala
+++ b/app/models/cache/Answers.scala
@@ -17,7 +17,7 @@
 package models.cache
 
 import forms.{AssociateUcr, MucrOptions, _}
-import models.UcrType.{Ducr, Mucr}
+import models.UcrType.{Ducr, DucrPart, Mucr}
 import models.cache.JourneyType.JourneyType
 import models.{Ucr, UcrBlock, UcrType}
 import play.api.libs.json.{Format, Json}
@@ -66,6 +66,8 @@ object ArrivalAnswers {
       ArrivalAnswers(consignmentReferences = Some(ConsignmentReferences(ConsignmentReferenceType.M, ucrBlock.ucr)))
     case Some(ucrBlock) if ucrBlock.ucrType.equals(Ducr.codeValue) =>
       ArrivalAnswers(consignmentReferences = Some(ConsignmentReferences(ConsignmentReferenceType.D, ucrBlock.ucr)))
+    case Some(ucrBlock) if ucrBlock.ucrType.equals(DucrPart.codeValue) =>
+      ArrivalAnswers(consignmentReferences = Some(ConsignmentReferences(ConsignmentReferenceType.DP, ucrBlock.ucr)))
     case _ =>
       ArrivalAnswers()
   }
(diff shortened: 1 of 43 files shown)


test/unit/models/UcrBlockSpec.scala
```diff
@@ -31,18 +31,18 @@ class UcrBlockSpec extends WordSpec with MustMatchers {
         val ucrBlock = UcrBlock(ucr = validDucr, ucrType = Ducr.codeValue)
```
SteveSugden, line 31: Why have you changed this test.  UcrBlock hasn't changed

PatrykRudnicki: Good catch, it was when I tried to remove `ucrPartNo` I renamed fullUcr to ucr and didn't change it back. Thanks

PatrykRudnicki: but this confirm what we discussed yesterday, that ucr field contains the ducr with part

PatrykRudnicki: I pushed a fix in separate commit for it

app/models/cache/Answers.scala
```diff
@@ -66,6 +66,8 @@ object ArrivalAnswers {
       ArrivalAnswers(consignmentReferences = Some(ConsignmentReferences(ConsignmentReferenceType.M, ucrBlock.ucr)))
     case Some(ucrBlock) if ucrBlock.ucrType.equals(Ducr.codeValue) =>
       ArrivalAnswers(consignmentReferences = Some(ConsignmentReferences(ConsignmentReferenceType.D, ucrBlock.ucr)))
+    case Some(ucrBlock) if ucrBlock.ucrType.equals(DucrPart.codeValue) =>
+      ArrivalAnswers(consignmentReferences = Some(ConsignmentReferences(ConsignmentReferenceType.DP, ucrBlock.ucr)))
```
SteveSugden, line 70: Would all of these DUCR Part answers be safer if they were using the ucrBlock.fullUcr method?    I know they currently return the same thing but if we decide to change the UcrBlock implementation to store the partId separately then this code here won't need to change.

Applies to all changes to this file

app/models/viewmodels/notificationspage/NotificationPageSingleElementFactory.scala
```diff
@@ -34,14 +34,16 @@ class NotificationPageSingleElementFactory @Inject()(responseConverterProvider:
 
   def build(submission: Submission)(implicit messages: Messages): NotificationsPageSingleElement =
     submission.actionType match {
-      case Arrival | RetrospectiveArrival | Departure | DucrDisassociation | MucrAssociation | MucrDisassociation | ShutMucr =>
+      case Arrival | RetrospectiveArrival | Departure | DucrDisassociation | DucrPartDisassociation | MucrAssociation | MucrDisassociation |
+          ShutMucr =>
         buildForRequest(submission)
-      case DucrAssociation => buildForDucrAssociation(submission)
+      case DucrAssociation     => buildForDucrAssociation(submission)
```
SteveSugden, line 40: Why is this different to ext movements?  There both DucrAssociation and DucrPartAssociation -> buildForDucrAssociation()

app/models/cache/Answers.scala
PatrykRudnicki: Yeah, pushed added this to review fixes commit. Good catch

app/models/viewmodels/notificationspage/NotificationPageSingleElementFactory.scala
PatrykRudnicki: This is the bug in the External Movements service, I'll create a fix for it there after I merge internal

SteveSugden: OK.  Great!

PR #197, opened 2020-08-04
CEDS-2640 Change content for DUCR Parts

PatrykRudnicki:


conf/messages.en
```diff
@@ -77,6 +77,7 @@ ducrPartDetails.ducrPartId.hint = This can be up to 3 digits. For example, 123.
 ducrPartDetails.ducrPartId.error = Enter a DUCR Part ID in the correct format
 
 consignmentReferences.reference.ducr = Declaration Consignment Reference (DUCR)
+consignmentReferences.reference.ducrPart = DUCR Part
```
SteveSugden, line 80: Is this text correct?  (DUCR Part).    Was there not some discussion that it should be similar to DUCR and MUCR?  Maybe "Declaration Consignment Reference Part (DUCR Part)"?

PatrykRudnicki: Don't know. I took the value from external service

PR #199, opened 2020-08-06
CEDS-2360 -fix for accessible-autocomplete

SteveSugden:
Fix for issue with accessible-autocomplete which prevents submission of empty values
https://github.com/alphagov/accessible-autocomplete/issues/432

Adapted from this fix in the amls-frontend project
https://github.com/hmrc/amls-frontend/pull/1457/commits/974839e999c06083abafb1a402c3ea30fd4b57c6


app/views/components/autocomplete/input_accessible_autocomplete.scala.html
```diff
@@ -40,6 +40,8 @@
 </div>
 
 <script src="/customs-exports-internal/assets/lib/accessible-autocomplete/dist/accessible-autocomplete.min.js" type="text/javascript"></script>
+<script src="/customs-exports-internal/assets/javascripts/jquery-3.5.1.slim.min.js" type="text/javascript"></script>
+<script src="/customs-exports-internal/assets/javascripts/auto-complete-fix.js" type="text/javascript"></script>
```
MaciejRewera, line 44: Didn't think about this place. But now I cannot imagine a better one 👍

app/assets/javascripts/auto-complete-fix.js
```diff
@@ -0,0 +1,62 @@
...
+    $('[data-gov-select-autocomplete]').each(function() {
+
+        var selectFieldName = $(this).attr('id').replace('[', '\\[').replace(']', '\\]');
+        var nonSelectFieldName = selectFieldName.replace('-select','');
+
+        var selectField = $('#' + selectFieldName)
+        var nonSelectField = $('#' + nonSelectFieldName)
+
+        nonSelectField.keydown(function(e) {
+            if (e.keyCode === 13 && $(this).val() === '') {
```
MaciejRewera, line 21: Nit: Might be worth assigning `13` to a constant.

SteveSugden:
I've removed some of the "redundant" code from the "fix" script.   Not sure what that code was supposed to do exactly but it seems to work without it.

PR #210, opened 2020-09-21
CEDS-2688 Upgrade Scala version to 2.12.12

MaciejRewera:


build.sbt
```diff
@@ -75,4 +77,15 @@ lazy val scoverageSettings: Seq[Setting[_]] = Seq(
   coverageFailOnMinimum := true,
   coverageHighlighting := true,
   parallelExecution in Test := false
-)
\ No newline at end of file
+)
+
+lazy val silencerSettings: Seq[Setting[_]] = {
```
SteveSugden, line 82: Are these the settings we need @PatrykRudnicki to get rid of our warnings?

SteveSugden (approved):
I'm no export, but LGTM

build.sbt
MaciejRewera: I've used them to silence dozens of warning coming from twirl templates and `.routes` files.

PatrykRudnicki: Yes, it will reduce amount of warnings

app/models/cache/Cache.scala
```diff
@@ -27,20 +27,6 @@ case class Cache(providerId: String, answers: Option[Answers], queryUcr: Option[
 }
 
 object Cache {
-  implicit private val formatInstant: OFormat[Instant] = new OFormat[Instant] {
```
PatrykRudnicki, line 30: Are you sure this is not required?

I remember that we had a problem with Instant, but not sure what was the problem.

PatrykRudnicki (approved):
Only worried about Instant format, but if everything works like excepted and ATs passed that should be okay

app/models/cache/Cache.scala
MaciejRewera: All tests passed (ATs too). Might not be an issue anymore.
Removed that after being pointed out by the compiler as unused.

PR #211, opened 2020-09-28
CEDS-2659 Upgrade SBT to 1.3.6

MaciejRewera:
This version seems to be the last reliable version.
Higher versions contain some issues with importing a project. That would
need additional steps and workarounds to make it work.


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=0.13.18
+sbt.version=1.3.6
```
PatrykRudnicki, line 1: The same like in other services. 1.3.13

PatrykRudnicki (approved):

PR #212, opened 2020-09-29
CEDS-2689 Add hidden text with context for movements

janeturner:
- Create option for visually hidden text on link component
- Use link with hidden text on submissions page

### DIFF
--- a/app/views/view_submissions.scala.html
+++ b/app/views/view_submissions.scala.html
@@ -37,7 +37,7 @@
   } else {
     @link(
       content = components.elements_in_separate_lines(submission.ucrBlocks.map(_.fullUcr)),
-      textHidden = Some(messages("submissions.hidden.text", submission.ucrBlocks.map(_.fullUcr).mkString)),
+      textHidden = Some(messages("submissions.hidden.text", submission.ucrBlocks.map(_.fullUcr).mkString(", "))),
       href = routes.ViewNotificationsController.listOfNotifications(submission.conversationId)
     )
   }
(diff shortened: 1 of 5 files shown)


app/views/view_submissions.scala.html
```diff
@@ -37,6 +37,7 @@
   } else {
     @link(
       content = components.elements_in_separate_lines(submission.ucrBlocks.map(_.fullUcr)),
+      textHidden = Some(messages("submissions.hidden.text", submission.ucrBlocks.map(_.fullUcr).mkString)),
```
MaciejRewera, line 40: This might not be what you want. Calling `mkString` without any parameters results in glue-ing the elements together. You might rather want to use something like: `.mkString(", ")` to separate the elements with comma.

test/unit/views/ViewSubmissionsViewSpec.scala
```diff
@@ -140,17 +140,17 @@ class ViewSubmissionsViewSpec extends ViewSpec with Injector {
       val secondDataRowElements = pageWithData.selectFirst(".govuk-table__body .govuk-table__row:nth-child(2)")
       val thirdDataRowElements = pageWithData.selectFirst(".govuk-table__body .govuk-table__row:nth-child(3)")
 
-      firstDataRowElements.selectFirst(".ucr").text() mustBe validMucr
+      firstDataRowElements.selectFirst(".ucr").text() mustBe s"$validMucr ${messages("submissions.hidden.text", validMucr)}"
```
MaciejRewera, line 143: It might be worth having a test with multiple UCRs. So you can test the `mkString` method.

MaciejRewera (approved):

PR #220, opened 2020-12-14
CEDS-2771 Validate MUCR length is <36 characters long

PauloPortugal:
Recently we have encountered the issue with following MUCR:

   `GB/82F9-0N2F6500040010TO120P0A300689` (36 characters long)

It complies with our regex but it was over 35 characters long and ILE
rejected for being too long.
After speaking with the QAs we have determined that the regex requires
an overal length validation.

The approach taken was not to overload the existing regex, it is already
too complex to read and instead add a separate validation on it's
length.


MaciejRewera (approved):
LGTM

PR #226, opened 2021-02-16
CEDS-2965 Set Root Url to redirect to home page

janeturner:
Create RootController to redirect to Find Consignment page.

### DIFF
--- a/app/controllers/RootController.scala
+++ b/app/controllers/RootController.scala
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2021 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


app/controllers/RootController.scala
```diff
@@ -0,0 +1,30 @@
...
+import play.api.i18n.I18nSupport
+import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
+import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendController
+
+import javax.inject.Inject
+
+class RootController @Inject()(mcc: MessagesControllerComponents) extends FrontendController(mcc) with I18nSupport {
+
+  def displayPage(): Action[AnyContent] = Action {
+    Redirect(controllers.ileQuery.routes.FindConsignmentController.displayQueryForm())
```
glidester, line 28: I know this is correct as per the ticket, but I have a feeling it should point to the 'choice' page (that redirects user to FindConsignment if providerId is missing). Let me check this with Beth.

glidester: No, the ticket is correct! :laughing:

glidester (approved):

PR #234, opened 2021-08-23
[CEDS-3315]: Removed remaining references to play-ui lib. Removed dep…

don-mac:
PR needs discussion, seeing intermittent failures in some of the smoke tests, related to failure to parse date and time form fields for Arrivals test case.

Removed remaining references to play-ui lib. 
Replicated components and templates in the project.
Replaced deprecated use of 'in' scope operator with slash notation in build.sbt


app/config/play_ui/AssetsConfig.scala
```diff
@@ -0,0 +1,26 @@
...
+
+package config.play_ui
+
+import javax.inject.Inject
+import play.api.Configuration
+
+class AssetsConfig @Inject()(configuration: Configuration) {
+  private val assetsUrl = configuration.get[String]("assets.url")
+  private val assetsVersion = configuration.get[String]("assets.version")
+  val assetsPrefix: String = assetsUrl + assetsVersion
```
glidester, line 25: Hi Don. Bit confused, seems like you've added in assets-frontend stuff that should be being removed as a dependency no?

don-mac: Unfortunately true, there were no direct equivalent for the majority of the components from play-ui in the play-frontend-hmrc library and this was how CEDS-3314 had been addressed.

PR #263, opened 2023-05-16
CEDS-4532 Update reads and writes for Cache

magicaljellybeans:
Add defensive reads so that updated fields is read as a string or Date

Ensure all new records are written with a Date to the updated field

### DIFF
--- a/app/models/cache/Cache.scala
+++ b/app/models/cache/Cache.scala
@@ -17,7 +17,9 @@
 package models.cache
 
 import models.{now, UcrBlock}
+import play.api.libs.functional.syntax.toFunctionalBuilderOps
 import play.api.libs.json._
+import uk.gov.hmrc.mongo.play.json.formats.MongoJavatimeFormats
 
 import java.time.Instant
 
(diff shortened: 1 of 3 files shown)


app/models/cache/Cache.scala
```diff
@@ -26,7 +28,24 @@ case class Cache(providerId: String, answers: Option[Answers], queryUcr: Option[
 }
 
 object Cache {
-  implicit val format: OFormat[Cache] = Json.format[Cache]
+  private val stringDateReads: Reads[Instant] = implicitly[Reads[String]].map(Instant.parse)
+
+  private val mongoDateReads: Reads[Instant] = MongoJavatimeFormats.instantReads
+
+  private val instantReads: Reads[Instant] = stringDateReads orElse mongoDateReads
```
glidester, line 35: Should this not try to read as a mongo date first (the correct format) and then only try string format if mongo format fails? :thinking:

magicaljellybeans: It will work as is but all new records will be mongo format so probably worth having that first

glidester (approved):

PR #285, opened 2024-04-03
CEDS-5634 Enable prefixed ROE codes to be parsed and displayed

magicaljellybeans:


app/models/viewmodels/decoder/ROECode.scala
```diff
@@ -26,47 +26,48 @@ import play.api.libs.json.{Format, JsResult, JsString, JsSuccess, JsValue}
  * @param code the code value
  * @param messageKey messages key with related description
  */
-sealed abstract class ROECode(override val code: String, override val messageKey: String, val priority: Int)
+sealed class ROECode(override val code: String, override val messageKey: String, val priority: Int, val prefixed: Boolean = false)
     extends Ordered[ROECode] with CodeWithMessageKey {
 
+  def displayCode: String = if (prefixed) "H" + code else code
```
glidester, line 32: Should replace "H" with `prelodgedPrefix` val instead

glidester (changes requested):

